PR #12, opened 2024-10-11
Add pyramids on cpu based on Jordao's code

ieivanov:
This PR largely copies https://github.com/royerlab/imgproc-monorepo/blob/main/impp/impp/pyramid.py expect that pyramids are implemented in CPU rather than GPU as stitches of OPS datasets do not easily fit in GPU memory. We should discuss how we can reduce down to one version of pyramiding functionality - maybe move that to `iohub`?

TODO:
- [x] Remove slurmkit dependency
- [x] Add pyramind to biahub CLI group

cc @ziw-liu @JoOkuma

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
 
 # list package dependencies here
 dependencies = [
-  "iohub[tensorstore]>=0.3.0a2,<0.4",
+  "iohub[tensorstore]>=0.3.0a5,<0.4",
   "stitch @ git+https://github.com/ahillsley/stitching@jen",
   "matplotlib",
   "napari",


Soorya19Pradeep:
Hello @ieivanov . I am trying to switch to this branch and pip install biahub in the conda environment I created. I am getting the following error:

> The conflict is caused by:
>     biahub 0.1.0 depends on iohub==0.2.0a1
>     waveorder 2.2.2.dev1378+g8315e00 depends on iohub==0.1.0
> 
> To fix this you could try to:
> 1. loosen the range of package versions you've specified
> 2. remove package versions to allow pip to attempt to solve the dependency conflict

Can you please help? Thanks!

ieivanov:
@Soorya19Pradeep I think you should be able to install `biahub` now. You'll also need to install slurmkit for this PR which is no longer a biahub dependency.

biahub/pyramid.py
```diff
@@ -0,0 +1,149 @@
...
+                int(round(current_scale[i] / previous_scale[i]))
+                for i in range(len(current_scale))
+            ]
+
+            downsampled = ts.downsample(
+                previous_level, downsample_factors=downsample_factors, method=method
+            )
+
+            target_store = dataset[str(level)].tensorstore()
+            target_store[:].write(downsampled[:].read().result()).result()
```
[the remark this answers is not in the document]
ziw-liu: See also https://github.com/google/tensorstore/issues/223

biahub/pyramid.py
```diff
@@ -0,0 +1,149 @@
...
+)
+
+
+def pyramid(fov_path: Path, levels: int, method: str) -> None:
+    """
+    Create pyramid levels for a single field of view using tensorstore downsampling.
+
+    This function uses cascade downsampling, where each level is downsampled from
+    the previous level rather than from level 0. This avoids aliasing artifacts
+    and chunk boundary issues that occur with large downsample factors.
```
ieivanov: Could this function move into `iohub`, say as a method of `Position`?

ieivanov:
On second thought and as folks pointed out, I agree that it doesn't make sense to convert floats to ints. It may make sense to convert the float32 data to float16 during pyramiding. For now, I think it's OK to keep all pyramid levels at float32, matching level `0`.

biahub/pyramid.py
```diff
@@ -0,0 +1,197 @@
...
+
+            current_scale = dataset.get_effective_scale(str(level))
+            previous_scale = dataset.get_effective_scale(str(level - 1))
+            downsample_factors = [
+                int(round(current_scale[i] / previous_scale[i]))
+                for i in range(len(current_scale))
+            ]
+
+            target_store = dataset[str(level)].tensorstore()
+            _write_ts_downsampled(
```
ieivanov: Let's switch to using the new `Position.compute_pyramid` method.

ieivanov:
@srivarra let's refactor this code such that it can use the latest changes in https://github.com/czbiohub-sf/iohub/pull/348

PR #25, opened 2024-12-01
fix to channel names during concat

edyoshikun:
Currently the concatenate CLI does not work if we pass the paths separatedly (i.e `[store.zarr/0/0/0,store.zarr/0/0/1,store.zarr/0/0/2]` and matching `channel names= [[`channel_1', 'channel_2'],[`channel_1', 'channel_2'],[`channel_1', 'channel_2']]`. This will result in an output store with 6 channel names repeated. 

This fixes that by only appending the channel names that don't exist.


ieivanov (approved):
Thanks @edyoshikun! I created #59 that we can address later. I'll merge this for now. I have a feeling that `concatenate` allows for too much flexibility in its inputs and can be hard to manage. Let's see, we may chose to reign it back a little in the future.

PR #51, opened 2025-03-01
Only raise error of different shapes if dataset is no going to be cropped

tayllatheodoro:
The code in main is raising an error if the files to be concatenated are not in the same shape, but for mantis analysis, the shape of the input will be different, and we crop before concatenating. So here, I just added a check the user passed values to crop the dataset; if they did, the error is not raised.


edyoshikun:
addressed in #25

Issue #52, opened 2025-03-06
Restructure biahub


ieivanov:
Here I'm suggesting a restructure of the `biahub` repo to make it easier to understand, browse, and contribute to.

### Make code easier to find

`biahub` currently has roughly the following structure:

```
biahub/
    analysis/
        deskew.py
        deconvolve.py
        ...
    cli/
        deskew.py
        deconvolve.py
        stabilize.py
        main.py
        ...
    tests/
```

The analysis module is a remnant from `shrimpy` where we had `acquisition` and analysis `modules`. Currently, it's confusing and hard to know what functions go under `analysis` vs `cli`. I suggest removing the `analysis` and `cli` modules to achieve a structure like this:

```
biahub/
    deskew.py
    deconvolve.py
    stabilize.py
    cli.py
    ...
    tests/
```

The new `deskew.py` module will combine functions from `analysis/deskew.py` and `cli/deskew.py` and define a `deskew_cli()` function which will be imported in `cli.py`, which will have the following structure:

```python
import click

from biahub.deskew import deskew_cli
...

@click.group()
def main()

main.add_command(deskew_cli)
...
```

### Make example config files easier to find

Our example configs, which are essential to find for any new user and buried in

```
biahub/
    analysis/
        settings/
            example_deskew_settings.yml
            example_deconvolve_settings.yml (actually currently missing)
            ...
```

We should instead move them to

```
biahub/
    deskew.py
    deconvolve.py
    stabilize.py
    cli.py
    tests/
settings/
    example_deskew_settings.yml
    example_deconvolve_settings.yml
```

@mattersoflight @edyoshikun @talonchandler @ziw-liu Let me know what you think about this plan. It may be helpful to merge large open PRs before this restructure, but I don't think that should be a blocker.

PR #61, opened 2025-03-17
Restructure biahub

ieivanov:
Restructure repo as discussed in #52


mattersoflight:
@tayllatheodoro (cc: @ieivanov), please run an integration test with https://github.com/czbiohub-sf/mantis-analysis-template. Please report here if any of the steps don't work as expected.

biahub/deconvolve.py
```diff
@@ -22,7 +24,46 @@
 from biahub.cli.utils import estimate_resources, get_output_paths, yaml_to_model
 
 
-@click.command()
+def compute_tranfser_function(
```
mattersoflight, line 27: @talonchandler @ieivanov Is there a call in waveorder to compute the transfer function? If yes, it is best to use library call.

mattersoflight: If there is a need for this method - please add a comment.

biahub/estimate_stabilization.py
```diff
@@ -13,7 +13,7 @@
 from pystackreg import StackReg
```
mattersoflight, line 13: Instead of using StackReg, I suggest implementing or reusing the 3D phase cross correlation that @edyoshikun may have implemented as part of autotracker algorithm. 3D phase cross correlation should work well for bead images.

mattersoflight: Is `ants` useful here?

mattersoflight (changes requested):
@ieivanov having methods and the CLI in the same file does make code quite readable. Most methods have docstrings, but a few do not. 

The logic and algorithms for registration and stabilization steps require an iteration. Let's brainstorm tomorrow or Thursday.

biahub/deconvolve.py
talonchandler: I 100% agree that there's some consolidation that can happen here, and I have this top of mind (especially because I'm reconstructing mantis and other datasets lately). 

I propose we complete this refactor, complete the waveorder-recorder merge, then I will lead a reconstruction refactor.   

IMO no need to block this PR on this point.

mattersoflight: That sounds good, @talonchandler. Please start an issue to remind yourself to fix this later.

talonchandler: [New issue here](https://github.com/czbiohub-sf/biahub/issues/63). Resolving this subdiscussion.

talonchandler (commented):
Just ran a very basic deconvolution integration test on this branch, and all worked as expected.

My only minor suggestion is to consider renaming `AnalysisSettings.py` to `analysis_settings.py` or just `settings.py`. Right now `AnalysisSettings.py` sticks out as the only CamelCase file.

biahub/estimate_stabilization.py
tayllatheodoro: I'm working on this in on this PR https://github.com/czbiohub-sf/biahub/pull/69, but results are not great for now

tayllatheodoro (approved):
The pipeline ran for all steps

PR #88, opened 2025-04-30
CLI for estimating XYZ crop parameters

ieivanov:
TODO:
- [x] Should be merged after #98 
- [x] Make sure all example test settings files are tested

This PR adds `biahub estiamte-crop` which finds the non-zero volume in the dataset across multiple channels (e.g. phase and fluorescence)

Our processing pipeline includes XYZ stabilization. This utility finds the largest overlapping volume between all timepoints.

The XY image size in phase is larger than the microscope aperture. `estimate-crop` allows us to provide a radius of this circular aperture (input as fraction of image width) that will be cropped out such that we end up with rectangular volume.

This implementation is still a bit hardcoded to our workflow and could be modified to be a bit more generic.


tayllatheodoro:
@ieivanov Hii ivan, can you check the code here?

PR #29, opened 2024-12-11
Add `process-with-config` CLI

edyoshikun:
This PR is inspired in the work from MONAI with composing transforms. I had to implement this for some quick normalization and binning. Similar to the current work in `cp_neurmast` branch. One can apply these functions in a sequence. 

Some major assumptions for now:
- the functions don't change the output much
- it's hardcoded that binning is supported to change the output shape
- some kwargs won't be able to be passed if they cannot be parsed by yml

PR #93, opened 2025-05-22
Add SLURM-enabled `biahub track` CLI for Ultrack-based cell tracking

tayllatheodoro:
## PR Summary: Add SLURM-enabled `biahub track` CLI for Ultrack-based cell tracking

This PR introduces a modular, SLURM-compatible command-line interface to perform 2D or 3D time-lapse cell tracking using the [Ultrack](https://github.com/royerlab/ultrack) library. It supports configurable preprocessing, per-channel pipelines, blank-frame handling, flexible Z-slice selection, and efficient job submission via Submitit.

### New Features

**Tracking CLI (`biahub track`):**
- Accepts a YAML configuration describing:
  - Input datasets (OME-Zarr)
  - Per-channel preprocessing pipelines
  - Tracking parameters for segmentation, linking, and global optimization
- Supports both **2D (Z-projected)** and **3D volumetric** tracking
- Submits one SLURM job per FOV using Submitit
- Includes automatic resource estimation and per-position logging

**Z-slice Selection Enhancements:**
- New `resolve_z_slice()` utility:
  - `(start, stop)` range
  - `(-1, -1)` for central slicing
  - `None` for full Z-range
- Ensures that at least one Z-slice is always selected

**Blank Frame Filling:**
- Handles empty timepoints using a CSV (`blank_frames_path`)
- Replaces invalid frames with nearest valid neighbors

**Configurable Preprocessing:**
- Supports `ProcessingInputChannel` with:
  - Input channels
  - Function names (`np.mean`, `ultrack.imgproc.normalize`, etc.)
  - Function kwargs and `per_timepoint` flag
- Functions are dynamically resolved and validated against a safe list

### Infrastructure & Utility Additions

- `update_model()` to merge nested Pydantic tracking configs
- `estimate_resources()` extended to support `max_num_cpus`
- `get_empty_frame_indices()` utility for identifying blank 3D frames
- `track_cli` added to the main CLI dispatcher (`main.py`)

### Developer Utilities

- A debugging script is available for interactive tracking parameter finetuning and Napari-based visualization:  
  [`scripts/debug_track_config_parameters.py`](scripts/debug_track_config_parameters.py)

---

### Example Usage

Tracking with Z-projected channels using configuration from `config_tracking.yml`:

```bash
biahub track \
  --output_dirpath output.zarr \
  --config_filepath config_tracking.yml

### CODE REVIEW

File: pyproject.toml [hunk 1]
```diff
@@ -42,6 +42,8 @@ dependencies = [
   "napari-psf-analysis",
   "imageio-ffmpeg",
   "dask[array]",
+  "toml",
+  "ultrack @ git+https://github.com/royerlab/ultrack.git"
```
JoOkuma, line 46: maybe add a git hash here to make sure you're always using the same ultrack version?

biahub/track.py
```diff
@@ -0,0 +1,656 @@
...
+    default_config = MainConfig()
+
+    tracking_cfg = update_model(default_config, tracking_data)
+
+    # Get the shape of the data
+    with open_ome_zarr(input_position_dirpaths[0]) as dataset:
+        T, C, Z, Y, X = dataset.data.shape
+
+    # Estimate resources
+    num_cpus, gb_ram_per_cpu = estimate_resources(shape=[T, C, Z, Y, X], ram_multiplier=16)
```
JoOkuma, line 597: Shouldn't ultrack's number of workers be a function of the number of CPUs per job?

And the number of parallel processes for ultrack (expect solving) scales with `T`, but this function requests resources in the order of `C x T`.

biahub/track.py
```diff
@@ -0,0 +1,656 @@
...
+      analysis.
+
+    Notes:
+    - The function modifies `tracking_config` to set the working directory for results storage.
+    - The function also saves the tracking configuration in a TOML file for reproducibility.
+    """
+
+    cfg = tracking_config
+
+    cfg.data_config.working_dir = databaset_path
```
JoOkuma, line 312: Using a shared storage as the working directory with the default `SQLite` backend can be a bit problematic.
The shared storage is often not fast enough to keep up with the multiple workers, and the database might get locked.
I suggest using something like `/tmp/{hash}` (or even `/dev/shm/{hash}`) and move the `data.db` and `metadata.toml`, and clean it afterwards.

biahub/track.py
```diff
@@ -0,0 +1,656 @@
...
+        input_segmentation_path = input_segmentation_dirpaths / Path(*position_key)
+
+        label_dataset = open_ome_zarr(input_segmentation_path)
+        label_arr = label_dataset[0][:, 0, :, :, :]  # Shape (T, Z, Y, X)
+
+        # Cast to uint32 in case Cellpose saved as float32
+        label_arr = label_arr.astype(np.uint32)
+
+        click.echo("Converting labels to foreground and contour gradient map...")
+        foreground_mask, contour_gradient_map = labels_to_edges(label_arr, sigma=90)
```
JoOkuma, line 496: This is mainly a sanity check, but 90 is a very large value for `sigma`.

biahub/track.py
```diff
@@ -0,0 +1,656 @@
...
+    for channel in input_channels["virtual_stain"]:
+        data_dict[channel] = vs_dataset[0][:, channel_names.index(channel), z_slices]
+
+    # Apply virtual staining projection function
+    if vs_projection_function is not None:
+        click.echo(
+            f"Applying {vs_projection_function.function} projection to the virtual staining data..."
+        )
+
+        projection = eval(vs_projection_function.function)  # Convert string to function
```
JoOkuma, line 477: Nitpicking, because avoiding `eval` is kind of a good practice.
This could be a nice alternative, assuming it's always a numpy function.
```suggestion
        projection = getattr(np, vs_projection_function.function)  # Convert string to function
```

biahub/track.py
```diff
@@ -0,0 +1,656 @@
...
+            data_dict[input_array] = projection(data_dict[input_array], **kwargs)
+
+    # Preprocess to get the the foreground and multi-level contours
+    click.echo("Preprocessing...")
+    if input_segmentation_dirpaths is not None:
+        click.echo(f"Loading segmentation from: {input_segmentation_dirpaths}...")
+        input_segmentation_path = input_segmentation_dirpaths / Path(*position_key)
+
+        label_dataset = open_ome_zarr(input_segmentation_path)
+        label_arr = label_dataset[0][:, 0, :, :, :]  # Shape (T, Z, Y, X)
```
JoOkuma, line 490: This might be an interesting use of `dask.from_zarr` to avoid loading everything into memory if that's an issue.

biahub/track.py
```diff
@@ -0,0 +1,656 @@
...
+        "dtype": np.uint32,
+    }
+
+    create_empty_hcs_zarr(
+        store_path=output_dirpath, position_keys=[position_key], **output_metadata
+    )
+
+    # Load virtual staining data
+    for channel in input_channels["virtual_stain"]:
+        data_dict[channel] = vs_dataset[0][:, channel_names.index(channel), z_slices]
```
JoOkuma, line 469: Same as below, these are cases where I like using `dask`.
But I would check if your functions handle Dask nicely (`np.asarray` to avoid creating large compute graphs).

settings/example_track_settings.yml
```diff
@@ -0,0 +1,89 @@
...
+
+tracking_functions:
+# functions to obtain tracking input arrays
+  foreground:
+    # Function for foreground detection.
+    function: ultrack.imgproc.detect_foreground
+    input_arrays:
+      - nuclei_prediction
+    kwargs: # Add additional parameters as required.
+      sigma: 90  # Gaussian blur standard deviation for foreground detection.
```
JoOkuma, line 81: The same as above, I'm just double-checking if `90` is good. It's much larger than what we are used to.

JoOkuma (commented):
Hi @tayllatheodoro, very clean PR.
I left some minor comments regarding resource utilization and some sanity checks.

pyproject.toml
ziw-liu: What is needed from the main branch that is not released?

biahub/track.py
```diff
@@ -0,0 +1,694 @@
...
+    _check_nan_n_zeros,
+    create_empty_hcs_zarr,
+    estimate_resources,
+    update_model,
+    yaml_to_model,
+)
+from biahub.settings import ProcessingFunctions, TrackingSettings
+
+
+# Custom function
```
ziw-liu, line 39:
```suggestion
```

biahub/track.py
```diff
@@ -0,0 +1,694 @@
...
+    contour_kwargs = contour_func_details.kwargs if contour_func_details.kwargs else {}
+
+    contour_gradient_map = array_apply(
+        *contour_input_arrays, func=contour_function, **contour_kwargs
+    )
+
+    return foreground_mask, contour_gradient_map
+
+
+def ultrack(
```
ziw-liu, line 276: This may cause confusion if `ultrack`-the-package is also imported in the call site's scope.

ziw-liu: LSP and other static analysis tools can also get confused.

biahub/track.py
```diff
@@ -0,0 +1,694 @@
...
+    input_vs_path: Path,
+    output_dirpath: Path,
+    input_channels: Dict[str, Any],
+    vs_projection_function: ProcessingFunctions,
+    preprocessing_functions: Dict[str, ProcessingFunctions],
+    tracking_functions: Dict[str, ProcessingFunctions],
+    z_slice: tuple,
+    z_shape: int,
+    tracking_config: MainConfig,
+    input_segmentation_dirpaths: Path = None,
```
ziw-liu, line 349: This is not in the docstring.

biahub/track.py
```diff
@@ -0,0 +1,694 @@
...
+
+        # Cast to uint32 in case Cellpose saved as float32
+        label_arr = label_arr.astype(np.uint32)
+
+        click.echo("Converting labels to foreground and contour gradient map...")
+        foreground_mask, contour_gradient_map = labels_to_edges(label_arr, sigma=90)
+
+    else:
+
+        foreground_mask, contour_gradient_map = data_preprocessing(
```
ziw-liu, line 500: Maybe a question for @JoOkuma: Is it better to always do intensity-based segmentation even when cellpose masks are available to get more hypotheses?

biahub/track.py
```diff
@@ -0,0 +1,993 @@
...
+
+    fov = "_".join(position_key)
+    click.echo(f"Processing FOV: {fov.replace('_', '/')}")
+    # tracking input images
+    foreground_mask, contour_gradient_map = data_preprocessing(
+        position_key, input_images, z_slices, blank_frames_path
+    )
+
+    # Define path to save the tracking database and graph
+    filename = str(output_dirpath).split("/")[-1].split(".")[0]
```
ilan-theodoro, line 765: output_dirpath.stem

biahub/cli/utils.py
```diff
@@ -643,9 +662,39 @@ def _is_nested(lst):
 
 def _check_nan_n_zeros(input_array):
```
ieivanov, line 663: In this PR, we are entirely changing how this function works - it goes from returning a boolean to returning a mixed output. Is this required for tracking? We should make sure we maintain backward compatibility.

Instead of having three different outputs - boolean (2D), list[int] (3D), list[(int, list)], I would suggest always outputting list[(channel_idx, z_idx)] of the C/Z index that contain a blank 2D image. This will ensure compatibility with the old use case of `if _check_nan_n_zeros(arr)` - the list will be empty if there are no blank images. In case there are blank images, you'd return, for example `[(0, 0)]` (2D), `[(0, 0), (0, 1)]` (3D), or `[(0, 0), (1, 0)]` (4D).

biahub/settings.py
```diff
@@ -23,6 +23,26 @@ class MyBaseModel(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
 
+class ProcessingFunctions(MyBaseModel):
```
ieivanov, line 26: This PR seems to include parts of #29? If so, then #29 should be merged first.

biahub/settings.py
```diff
@@ -23,6 +23,26 @@ class MyBaseModel(BaseModel):
...
+
+class ProcessingInputChannel(MyBaseModel):
+    path: Union[str, List[str], None]
+    channels: Dict[str, List[ProcessingFunctions]]
+
+
+class TrackingSettings(MyBaseModel):
+    target_channel: str = "nuclei_prediction"
+    mode: Literal["2D", "3D"] = "2D"
+    z_slices: Optional[Tuple[int, int]] = None
```
ieivanov, line 41: I suggest calling this `z_range` (e.g. (0, 10)) to avoid confusion with slice objects created by `resolve_z_slice` later which would return `slice(0, 10)`

biahub/track.py
```diff
@@ -0,0 +1,993 @@
...
+    if n_slices % 2 == 0:
+        n_slices += 1
+    z_center = z_shape // 2
+    half_window = n_slices // 2
+    return slice(z_center - half_window, z_center + half_window + 1)
+
+
+def resolve_z_slice(z_slices: Tuple[int, int], z_shape: int, mode: str = "2D") -> slice:
+    """
+    Resolve the z-slice range based on user-defined input and imaging mode.
```
ieivanov, line 294: I think this function needs better documentation describing all use cases. `z_shape` should also be documented as the number of slices in the *input* dataset.

settings/example_track_settings.yml
```diff
@@ -0,0 +1,86 @@
...
+
+# Name of the target channel used to name the tracking output label channel.
+target_channel: nuclei_prediction
+
+# Input image configurations.
+# Each entry defines one dataset with its path and channels to process.
+input_images:
+
+  # First input: predicted nuclei and membrane channels.
+  - path: /path/to/virtual_staining.zarr/*/*/*
```
ieivanov, line 15: This field is redundant with the `input_position_dirpaths` argument of the tracking CLI. Paths to images should be provided only ones, during the CLI call

biahub/track.py
```diff
@@ -0,0 +1,993 @@
...
+    click.echo('Submitting SLURM jobs...')
+    jobs = []
+
+    with executor.batch():
+        for input_position_path in input_position_dirpaths:
+            position_key = input_position_path.parts[-3:]
+            job = executor.submit(
+                track_one_position,
+                position_key=position_key,
+                output_dirpath=output_dirpath,
```
ieivanov, line 929: Here you should iterate over iohub Position objects like we do in all other CLIs. This way you won't need to pass the path to the data using the tracking config file, related to this other comment: https://github.com/czbiohub-sf/biahub/pull/93#discussion_r2162570350

biahub/concatenate.py
```diff
@@ -348,15 +348,15 @@ def concatenate(
...
-    num_cpus, gb_ram_per_cpu = estimate_resources(shape=[T, C, Z, Y, X], ram_multiplier=16)
+    num_cpus, gb_ram_per_cpu = estimate_resources(shape=[T, C, Z, Y, X], ram_multiplier=16, max_num_cpus=16)
     # Prepare SLURM arguments
     slurm_args = {
         "slurm_job_name": "concatenate",
         "slurm_mem_per_cpu": f"{gb_ram_per_cpu}G",
         "slurm_cpus_per_task": num_cpus,
         "slurm_array_parallelism": 100,  # process up to 100 positions at a time
         "slurm_time": 60,
-        "slurm_partition": "cpu",
```
ieivanov, line 359: We should keep concatenate jobs as preempted

biahub/track.py
```diff
@@ -0,0 +1,993 @@
...
+    Example usage:
+
+    biahub track -i virtual_staining.zarr/*/*/* -o output.zarr -c config_tracking.yml -f blank_frames.csv
+
+    """
+
+    track(
+        output_dirpath=output_dirpath,
+        config_filepath=config_filepath,
+        input_position_dirpaths=input_position_dirpaths,
```
ieivanov, line 985: How do the `input_position_dirpaths` here work together with the image paths given in the tracking config file? From what I understand, the images provided in the config file are used to compute contours which are used to perform the tracking. Once we have tracks, they are overlayed onto the data in `input_position_dirpaths`. Is that right? Is the data pass with `input_position_dirpaths` modified in any way? Should the output of the tracking CLI be a CSV file that could be associated with any zarr store?

ieivanov: I see that tracking generated a CSV file and a zarr file with object labels. I'm guessing these are added as channels to the data in `input_position_dirpaths`? I suggest that tracking output only a zarr with the labels - these can then be concatenated to any other zarr store. This would lead to less data duplication

biahub/track.py
```diff
@@ -0,0 +1,993 @@
...
+        toml.dump(cfg.dict(by_alias=True), f)
+
+    return (
+        labels,
+        tracks_df,
+        graph,
+    )
+
+
+def run_pipeline(
```
ieivanov, line 404: I would suggest renaming to run_preprocessing_pipeline

biahub/track.py
```diff
@@ -0,0 +1,993 @@
...
+                if visualize:
+                    import napari
+
+                    viewer = napari.Viewer()
+                    viewer.add_image(data_dict[channel_name], name=channel_name)
+
+    return data_dict
+
+
+def read_data(
```
ieivanov, line 498: This function is used only once in `load_data`, I think we don't need it, i.e. reading the data doesn't need to be that refactored, especially since `shape` and `scale` are not used in `load_data`

biahub/track.py
```diff
@@ -0,0 +1,982 @@
...
+
+from biahub.cli.parsing import (
+    config_filepath,
+    local,
+    output_dirpath,
+    sbatch_filepath,
+    sbatch_to_submitit,
+)
+from biahub.cli.utils import (
+    create_empty_hcs_zarr,
```
ziw-liu, line 28: Should use the iohub version.

ieivanov (approved):
Great works, thanks @tayllatheodoro! Approving, let's merge this PR after #29

PR #94, opened 2025-05-24
Parallelized, low-RAM stitching with flexible blending

talonchandler:
@ahillsley and I realized that our existing stitching implementations are not easily parallelized, require a lot of RAM, and do not support flexible blending. We went back to the drawing board and revised our algorithm into:

1) For each output chunk, find all of the input FOVs that contribute. 
2) For each contributing input FOV, compute a weight map. 
3) Load the contributing FOVs, apply weights, sum, and write to the output zarr. 

This algorithm :
- runs in parallel---by default it sends jobs to SLURM, use the `-l` flag for local. 
- does not require much RAM (we never make a large array of zeros),
- supports flexible blending (average b=0, linear b=1, and progressively sharper S-curve b>1)

**PR Status:**
- [x] test on single-timepoint multi-channel fish from hummingbird (benchmark: 2 minutes for a P=14, T=1, C=4, Z=40, Y=2048, X=2048, blended stitch)
- [x] test on multi-timepoint multi-channel fish from hummingbird
- [x] test on multi-channel zebrafish from dragonfly (with flipping)
- [x] test on multi-channel OPS

### DIFF
--- a/biahub/estimate_stitch.py
+++ b/biahub/estimate_stitch.py
@@ -35,37 +35,33 @@ def extract_stage_position(
     stage_positions = plate_dataset.zattrs["Summary"]["StagePositions"]
     for stage_position in stage_positions:  # TODO: fail if this loop reaches the end
         if stage_position["Label"] == position_name:
+            # Initialize default values
+            xpos, ypos, zpos = 0, 0, 0
 
-            # Read XY positions
-            try:
-                xy_stage_name = stage_position["DefaultXYStage"]
-                if "DevicePositions" in stage_position.keys():
-                    for device in stage_position["DevicePositions"]:
-                        if device["Device"] == xy_stage_name:
-                            xpos, ypos = device["Position_um"]
-                            break
-                else:
-                    xpos, ypos = stage_position[xy_stage_name]
-            except KeyError:
-                xpos, ypos = 0, 0
-                pass
-
-            # Read Z positions
-
-            try:
-                xy_stage_name = stage_position["DefaultXYStage"]
-                z_stage_name = stage_position['DefaultZStage']
+            if "DevicePositions" in stage_position.keys():
+                # Handle DevicePositions case
+                xy_stage_name = stage_position.get("DefaultXYStage", "")
                 non_z_devices = {xy_stage_name}
-                if "DevicePositions" in stage_position.keys():
-                    zpos = 0
-                    for device in stage_position["DevicePositions"]:
-                        if device["Device"] not in non_z_devices:
-                            zpos += device["Position_um"][0]
-                else:
+
+                for device in stage_position["DevicePositions"]:
+                    if device["Device"] == xy_stage_name and xy_stage_name:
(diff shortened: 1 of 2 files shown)


talonchandler:
Initial trials with linear blending are encouraging. This is our first 2D tile grid with blending.  
Screenshot 2025-05-23 at 9 22 35 PM

talonchandler:
Also tested on a multi-fish multi-timepoint blended stitch. `biahub flip` supports post-stitching flips. Wall time was ~10 minutes. 
Screenshot 2025-05-29 at 12 26 12 PM

I have trialled this branch for the major zebrafish use cases, and I will use this branch for my upcoming shift-variant reconstruction use cases.

@ahillsley when you're back let's work together to take this across the finish line for OPS. I took an initial look, but I found that stage position metadata wasn't included in your datasets, and I think your scripts estimate the shifts tilt by tile. I can help with the integration work so that OPS can benefit from the blending, low-RAM, and parallelization available in this branch.

mattersoflight:
@ahillsley @talonchandler Sri can take this forward after he starts on August 4th!  We can figure out when he should tackle this PR and merge this into biahub for use across OPS, whole zebrafish volumes, and DaXi for deconvolutions.

biahub/stitch.py
```diff
@@ -1,440 +1,367 @@
...
-        if not isinstance(row_translation, tuple):
-            row_translation = (0, row_translation)
-        xy_output_shape, global_translation = get_stitch_output_shape(
-            n_rows, n_cols, sizeY, sizeX, col_translation, row_translation
-        )
-    else:
-        df = pd.DataFrame.from_dict(
-            total_translation, orient="index", columns=["shift-y", "shift-x"]
+        click.echo("\t\tBuilding weight maps")
+    w = np.power(distance_maps, blending_exponent, where=(distance_maps > 0))
```
srivarra, line 165: [question(if-minor)](https://pullpo.io/cc?l=question&d=if-minor) 
Is the where condition redundant? Won't the `distance_maps` array always contain values 0 or greater, and $0^{blending\\_exponent} = 0$?

biahub/stitch.py
```diff
@@ -1,440 +1,367 @@
...
+                    write_output_chunk,
+                    *job_args,
+                )
+            )
+
+    job_ids = [job.job_id for job in jobs]  # Access job IDs after batch submission
+
+    log_path = Path(slurm_out_path / "submitit_jobs_ids.log")
+    with log_path.open("w") as log_file:
+        log_file.write("\n".join(job_ids))
```
srivarra, line 364: [thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 
I see this pattern quite a bit throughout `biahub`, would it be beneficial to have a couple of classes or functions which we could reuse instead?

srivarra (changes requested):
This looks good overall. The many functions need documentation, and I have one minor suggestion plus a few questions.

PR #97, opened 2025-06-04
Parallelization of waveoder cli in biahub

tayllatheodoro:
- [ ] bring code from waveorder
- [ ] adapt to biahub style
- [ ] add example files
- [ ] test clis

### DIFF
--- a/biahub/apply_inverse_transfer_function.py
+++ b/biahub/apply_inverse_transfer_function.py
@@ -1,0 +1,132 @@
+import itertools
+import os
+import warnings
+from functools import partial
+from pathlib import Path
+from typing import Final
+
+import click
+import numpy as np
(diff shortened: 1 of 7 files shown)


talonchandler (changes requested):
Thanks for the initial pass @tayllatheodoro. I'm looking forward to having this complete. 

I will suggest we skip the duplication of these settings files. I automatically generate these in `waveorder`, and the format will be changing. Perhaps we mention these examples in the docstring and make it clear that these are `waveorder` wrappers?

talonchandler:
cc @edyoshikun who might be waiting on this to merge.

ieivanov (approved):
Thanks @talonchandler! This PR looks good, merging...

PR #100, opened 2025-06-11
Faster CI

### LINKED ISSUE #91: Faster style check in CI
Performing style check in CI doesn't need to install the biahub package (with torch and its CUDA friends) which takes minutes. See this [example](https://github.com/mehta-lab/VisCy/blob/2d57fd2c94a07b4f0cf11d3788728656fc950e57/.github/workflows/pr.yml#L7).

ieivanov:
Fixes #91. To be merged after #97 

In this PR:
- Style checks run before pytest; pytest requires successful style check
- Updated style dependency (black, flake8) versions
- Switched CI testing environment from `windows-latest` to `ubuntu-latest`
- Fixed bug in get_mean_z_positions: in the previous tests we were trying to forward fill a missing z_idx at t=0 which is ill defined. Now we no longer do a forward fill, instead only computing mean across accurately calculated z_idx's
- Added `sbatch_file` fixture which limits CPU utilization during testing
- Updated sbatch_to_submitit to allow configuring local jobs as well (not just SLURM jobs)
- Limited concatenate tests to use a `debug` cluster which means that they'll execute sequentially rather than in parallel to avoid problems with too many processes running simultaneously in the GitHub CI which often causes jobs to be terminated early.

### DIFF
--- a/biahub/tests/test_concatenate.py
+++ b/biahub/tests/test_concatenate.py
@@ -429,6 +429,7 @@ def test_concatenate_with_unique_positions(create_custom_plate, tmp_path, sbatch
     concatenate(
         settings=settings_unique,
         output_dirpath=output_path_unique,
+        sbatch_filepath=sbatch_file,
         local=True,
     )
 


ieivanov:
TODO: try adding back cpu-only torch install

biahub/tests/test_concatenate.py
```diff
@@ -387,7 +396,7 @@ def test_concatenate_with_mixed_slice_formats(create_custom_plate, tmp_path):
     )
 
 
-def test_concatenate_with_unique_positions(create_custom_plate, tmp_path):
+def test_concatenate_with_unique_positions(create_custom_plate, tmp_path, sbatch_file):
```
ziw-liu, line 399: Is this actually used?

ziw-liu: I wasn't sure if the argument in this test case gets used since there is only diff in the signature, not in the function body.

ieivanov: Ah, yes, good catch, thanks. Fixed.

ziw-liu (approved):

ziw-liu:
@ieivanov is this ready to merge?

PR #101, opened 2025-06-16
Updating the segmentation to use Cellpose SAM

edyoshikun:
This PR updates the segmentation pipline to use the Cellpose-SAM.


biahub/segment.py
```diff
@@ -83,13 +85,21 @@ def segment_data(
             )
             czyx_data[c_idx] = func(czyx_data[c_idx], **kwargs)
 
+        # Reorder channels based on channels_for_segmentation
+        cellpose_czyx = np.zeros(
+            (3, *czyx_data_to_segment.shape[1:]), dtype=czyx_data_to_segment.dtype
```
ziw-liu, line 90: Is this the output datatype? If it's not stored in the same as input, it should always be integers whose width is determined automatically by cellpose. For example storing uint32 in a float32 container is lossy.

ieivanov:
@edyoshikun are you happy with this PR? Is it ready to merge? I understand this introduces a breaking change in the segmentation config, which we can coordinate with @tayllatheodoro - I think currently segmentation is not part of the processing pipeline, through it probably should be.

biahub/segment.py
```diff
@@ -200,17 +233,17 @@ def segment_cli(
     )
 
     # Estimate resources
-    num_cpus, gb_ram_request = estimate_resources(shape=segmentation_shape, ram_multiplier=20)
+    num_cpus, gb_ram_request = estimate_resources(shape=segmentation_shape, ram_multiplier=10)
```
ieivanov, line 236: This seems like a lot of funny math to estimate the resources. Does `estimate_resources` not do everything you need? For example, it now allows for specifying the max number of CPUs.

biahub/segment.py
```diff
@@ -236,12 +269,12 @@ def segment_cli(
...
-                    segment_data,
-                    input_position_path,
-                    output_position_path,
+                    func=segment_data,
+                    input_position_path=input_position_path,
+                    output_position_path=output_position_path,
                     input_channel_indices=[list(range(C))],
                     output_channel_indices=[list(range(C_segment))],
-                    num_processes=np.min([20, int(num_cpus * 0.8)]),
+                    num_processes=np.min([slurm_array_parallelism, int(num_cpus * 0.8)]),
```
ieivanov, line 277: The number of processes here should be decoupled from the `slurm_array_parallelism`?

biahub/segment.py
ieivanov: The real solution here would be to save integer labels in a separate `labels` array in the zarr store. For now we save everything in one float32 array, which is not ideal. I'd suggest casting cellpose_czyx to float32 and leaving a note that we should fix that later.

biahub/segment.py
ieivanov: I'd suggest leaving reasonable defaults in the code and using an sbatch file to tune these parameters as needed for specific reconstructions or depending on the cluster usage

biahub/segment.py
edyoshikun: I can default it to uint32.

biahub/segment.py
edyoshikun: It's tricky because this depends on the input dataset.

edyoshikun:
I couldn't immediately figure out why when we run it locally, it doens't seem to find the GPU devices and defaults to cpu, which is super slow for cellpose.

ieivanov:
@edyoshikun let's discuss the resource allocation in the PR and then we can merge it if you're happy with it.

ieivanov:
@edyoshikun @allysonryan let's revisit if and when we should merge this branch

PR #104, opened 2025-07-03
Concatenate with OME-Zarr v0.5 and sharding

ziw-liu:
Needs https://github.com/czbiohub-sf/iohub/pull/311

### DIFF
--- a/biahub/concatenate.py
+++ b/biahub/concatenate.py
@@ -392,38 +392,39 @@ def concatenate(
     click.echo("Submitting SLURM jobs...")
     jobs = []
 
-    with executor.batch():
-        for i, (
-            input_position_path,
-            output_position_path,
-            input_channel_idx,
-            output_channel_idx,
-            zyx_slicing_params,
-        ) in enumerate(
-            zip(
-                all_data_paths,
-                output_position_paths_list,
-                input_channel_idx_list,
-                output_channel_idx_list,
-                all_slicing_params,
-            )
-        ):
-            # Create slicing parameters for this specific path
-            copy_n_paste_kwargs = {"zyx_slicing_params": zyx_slicing_params}
+    with submitit.helpers.clean_env():
+        with executor.batch():
+            for i, (
+                input_position_path,
+                output_position_path,
+                input_channel_idx,
+                output_channel_idx,
+                zyx_slicing_params,
+            ) in enumerate(
+                zip(
+                    all_data_paths,
+                    output_position_paths_list,
+                    input_channel_idx_list,
+                    output_channel_idx_list,
+                    all_slicing_params,
+                )
+            ):
+                # Create slicing parameters for this specific path
+                copy_n_paste_kwargs = {"zyx_slicing_params": zyx_slicing_params}
 
-            job = executor.submit(
-                process_single_position,
-                copy_n_paste,
-                input_position_path=input_position_path,
-                output_position_path=output_position_path,
-                input_channel_indices=input_channel_idx,
-                output_channel_indices=output_channel_idx,
-                input_time_indices=input_time_indices,
-                output_time_indices=list(range(len(input_time_indices))),
-                num_processes=int(slurm_args["slurm_cpus_per_task"]),
-                **copy_n_paste_kwargs,
-            )
-            jobs.append(job)
+                job = executor.submit(
+                    process_single_position,
+                    copy_n_paste,
+                    input_position_path=input_position_path,
+                    output_position_path=output_position_path,
+                    input_channel_indices=input_channel_idx,
+                    output_channel_indices=output_channel_idx,
+                    input_time_indices=input_time_indices,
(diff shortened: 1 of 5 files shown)


ziw-liu:
Ran into https://github.com/zarr-developers/zarr-python/issues/3221.

ziw-liu:
> As of [83cd243](https://github.com/czbiohub-sf/biahub/commit/83cd243d581d1832b430df1fc1c8d65ce9aa3e38) converting a 282 GB dataset (325 GB decompressed) took 7 minutes on 2 nodes with 64 CPUs each. As a reference converting a 65 GB dataset (183 GB decompressed) takes 2 minutes on 16 CPUs when using thread parallelism.

https://github.com/czbiohub-sf/biahub/pull/104/commits/606a0c4c7f010d4f1a6a638dbdf99f239d120c69 now taking about 2 minutes.

pyproject.toml
```diff
@@ -47,11 +47,16 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-# note that dev dependencies are only pinned to major versions
-segmentation = ["cellpose"]
-sharding = ["tensorstore"]
+segment = ["cellpose"]
+
+track = ["ultrack @ git+https://github.com/royerlab/ultrack.git@v0.6.3"]
```
ziw-liu, line 52: Is this any different than the one from PyPI?

ieivanov: Good question - I'm guessing no? @tayllatheodoro may know better

biahub/concatenate.py
```diff
@@ -238,22 +238,32 @@ def calculate_cropped_size(
 def concatenate(
     settings: ConcatenateSettings,
     output_dirpath: Path,
-    sbatch_filepath: str = None,
+    sbatch_filepath: str | None = None,
     local: bool = False,
+    block: bool = False,
```
ieivanov, line 243: What was the motivation for including the block parameter? Was it useful during testing?

ziw-liu: When running locally there isn't a good way to check if the jobs (processes) have finished. It is also useful for testing.

ieivanov:
Current plan is that this PR will be merged after https://github.com/czbiohub-sf/iohub/pull/301, updating the iohub dependency to the main branch.

srivarra (commented):
LGTM, I was able to run `biahub concatenate -c rechunk.yml -o test.zarr -sb sbatch.sh` on a dataset which hasn't been converted from OME-NGFF v0.4/Zarr V2 to OME-NGFF v0.5/Zarr V3 over here `/hpc/projects/intracellular_dashboard/organelle_dynamics/rerun/2025_04_15_A549_H2B_CAAX_ZIKV_DENV/2-assemble/zarr-v3`.

srivarra:
@ieivanov Should we wait for a new release candidate for `ultrack` so it supports Zarr V3? It's updated on the main branch and is the last dependency for `biahub` (it is optional) to get sorted out. Or should I pin it to main in the `pyproject.toml`?

edyoshikun:
https://github.com/royerlab/ultrack/issues/245

ieivanov:
@srivarra I agree that the right strategy for `ultrack` is to get a pre-release, as you've suggested in https://github.com/royerlab/ultrack/issues/245. Ping Jordao on Slack too, this should be relatively easy to do. A GitHub release that's not published on pypi will also be OK. As a fallback, we can depend on a specific commit in main - which will be pretty much equivalent to a pre-release tag.

@Soorya19Pradeep I think the right way to convert zarr stores from v2 to v3 is using the `biahub concatenate` CLI. The reason to use biahub for conversion is that we can parallelize the conversion of multiple FOVs across multiple nodes with SLURM. If we implemented something like that in iohub it would be slower as it uses the resources on one node only. (With iohub it's still possible to read a zar v2 store and save it as zarr v3 using the python API)

I agree that we should document this use case better by extending the `concatenate` CLI docstring in this PR and providing an `example_convert_zarr_v3_settings.yaml` file - it will be pretty similar to `example_concatenate_settings.yml` but will only contain paths to one v2 store in `concat_data_paths`. @srivarra can you please work on that? I agree that doing conversion using the `concatenate` CLI is a bit confusing. Do you think it's work creating a `convert` CLI which is pretty much an alias to `concatenate`, maybe requiring that you only provide a single zarr store?

@Soorya19Pradeep so far we've only tested the concatenation of zarr v2 stores. It's possible that there are bugs concatenating zarr v3 stores. We've also not tried concatenating stores with pyramid levels. You can work with Sri to debug these features.

I agree that concatenate should carry over the scale metadata (time scale, time unit, etc.) It's possible that you get an error if the two stores you are concatenating don't have the same time scale at which point the algorithm won't know which one to pick. But in either case, the time scale shouldn't go missing.

I think we've never tried transferring the omero metadata (contrast limits, default T and Z) during concatenation. @srivarra could you add these features?

PR #122, opened 2025-08-04
Fix SQLite concurrency by ensuring new database_path

tayllatheodoro:
Issue:
Parallel SLURM jobs were writing simultaneously to a shared SQLite database, causing concurrency errors (UNIQUE constraint failed: nodes.id).

Fix:

Create a unique database_path for each field of view (FOV).

Raise error if output_filepath exist
This ensures safe parallel execution without SQLite integrity errors.


biahub/track.py
```diff
@@ -720,13 +721,20 @@ def track_one_position(
 
     # Define path to save the tracking database and graph
     filename = output_dirpath.stem
-    databaset_path = output_dirpath.parent / f"{filename}_config_tracking" / f"{fov}"
-    os.makedirs(databaset_path, exist_ok=True)
+    database_path = output_dirpath.parent / f"{filename}_config_tracking" / f"{fov}"
+    if database_path.exists():
+        click.echo(f"Removing existing database at {database_path} to avoid SQLite conflicts.")
+        shutil.rmtree(database_path)
```
ieivanov, line 727: As we discussed, update code such that it does not call `shutil.rmtree`

PR #126, opened 2025-08-07
Check Slurm Job logs status

tayllatheodoro:
https://github.com/czbiohub-sf/biahub/issues/111

- [x] function to check logs dir
- [x] cli to check logs dir
- [ ] submitit a job after other clis 
Other options for implementation:

### DIFF
--- a/biahub/cli/slurm.py
+++ b/biahub/cli/slurm.py
@@ -1,12 +1,14 @@
-import submitit
-import click
 import os
-from biahub.cli.parsing import output_dirpath
-from tqdm import tqdm
-from pathlib import Path
-import pandas as pd
-from typing import Optional, Union
+
 from datetime import datetime
+from pathlib import Path
+from typing import Optional, Union
+
+import click
+import pandas as pd
+import submitit
+
+from tqdm import tqdm
 
 
 def wait_for_jobs_to_finish(jobs: list[submitit.Job]) -> None:
@@ -31,9 +33,10 @@ def wait_for_jobs_to_finish(jobs: list[submitit.Job]) -> None:
 
 
 def check_job_logs(
-        log_dir: Union[str, Path],
-        job_ids: Optional[list[Union[str, int]]] = None,
-        output_dir: Union[str, Path] = None) -> pd.DataFrame:
+    log_dir: Union[str, Path],
+    job_ids: Optional[list[Union[str, int]]] = None,
+    output_dir: Union[str, Path] = None,
+) -> pd.DataFrame:
     """
     Check SLURM logs for specific job IDs or all jobs in a folder,
     reporting statuses and saving a filtered CSV with non-successful jobs.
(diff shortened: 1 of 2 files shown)


biahub/cli/slurm.py
```diff
@@ -22,3 +30,136 @@ def wait_for_jobs_to_finish(jobs: list[submitit.Job]) -> None:
             pass  # as_completed polls every 10 seconds by default, so we don't need to do anything here
         except Exception as e:
             print(f"Job {job.job_id} failed with exception: {e}")
+
+
+def check_job_logs(
+    log_dir: Union[str, Path],
```
srivarra, line 36: Instead of `Union[str, Path]` I recommend either using use: `str | Path` or `PathLike` from `os`. `str | Path` is the recommended notation for Union types for Python 3.10+ (it's less to type).

`PathLike` is a nice protocol if you want a catch-all for everything that is well... "path like".

biahub/cli/slurm.py
```diff
@@ -22,3 +30,136 @@ def wait_for_jobs_to_finish(jobs: list[submitit.Job]) -> None:
...
+    job_ids: Optional[list[Union[str, int]]] = None,
```
srivarra, line 37: This type hint can also be simplified `list[str | int] | None = None`.

biahub/cli/slurm.py
```diff
@@ -22,3 +30,136 @@ def wait_for_jobs_to_finish(jobs: list[submitit.Job]) -> None:
...
+
+    if job_ids is None:
+        job_ids = sorted(
+            [
+                fname.name.replace("_log.out", "").replace("_log.err", "")
+                for fname in log_dir.glob("*_log.out")
+            ]
+        )
+
+    for job_id in tqdm(job_ids, desc="Analyzing SLURM logs"):
```
srivarra, line 67: Q: Is this slow enough where a progress meter makes sense?

biahub/cli/slurm.py
```diff
@@ -22,3 +30,136 @@ def wait_for_jobs_to_finish(jobs: list[submitit.Job]) -> None:
...
+                        logs_combined += log_err.read_text().lower()
+
+                    if (
+                        "job completed successfully" in logs_combined
+                        or "exiting after successful completion" in logs_combined
+                    ):
+                        status = "SUCCESS"
+                    else:
+                        status = "FAILED"
+                except Exception as e:
```
srivarra, line 97: Bare exceptions can be uninformative. Do we know what kinds of errors would show up here? Would a `FileNotFound` make sense, or maybe an `OSError`?

Here's a list of default [Python exceptions][py-errors].

[py-errors]: https://docs.python.org/3/library/exceptions.html

srivarra (changes requested):
Just some minor stuff about typehints mainly and a question about `tqdm` and when the log file can't be read.

PR #129, opened 2025-08-08
Add execution time prints

tayllatheodoro:
https://github.com/czbiohub-sf/biahub/issues/128


biahub/cli/parsing.py
```diff
@@ -11,6 +13,18 @@
...
+def timed(fn):
+    @wraps(fn)
+    def wrapper(*args, **kwargs):
+        start = time.time()
+        result = fn(*args, **kwargs)
+        end = time.time()
+        print(f"Execution Time: {end - start:.2f} seconds")
+        return result
+
+    return wrapper
```
srivarra, line 25: A couple of suggestions:

1. **Use `time.perf_counter()` instead of `time.time()`**: `time.time()` doesn't account for system changes (like daylight savings) and is better suited for timestamps. For timing, `perf_counter()` is more reliable.

2. **Improve execution time formatting**: Suppose the time it took for some function was `123123.123456` seconds. That's a bit hard to see and translate into something we, mere mortals can perceive. I recommend using the [`humanize`][humanize-gh] package for human-readable output:

```python
import humanize
td = humanize.precisedelta(
    dt.timedelta(seconds=123123.123456), minimum_unit="microseconds"
)
print(td)
# Output: 1 day, 10 hours, 12 minutes, 3 seconds, 123 milliseconds and 456 microseconds
```

[humanize-gh]: https://github.com/python-humanize/humanize

srivarra (changes requested):

biahub/cli/parsing.py
ieivanov: @srivarra can you please implement these changes and get @tayllatheodoro 's review?

ieivanov:
I think that looks good. I assume you guys have tested it? Are you concerned that when importing the function through a python API it will always print the execution time? I think thats OK? If you are happy with it, please add this decorator to other functions in the CLI library and let's merge this PR

PR #130, opened 2025-08-08
Improve CLI Startup Performance with Lazy Command Loading

tayllatheodoro:
###Overview
This PR introduces lazy loading of subcommands in the biahub CLI to significantly reduce startup time. Instead of importing all subcommands and their dependencies at once, each command is now loaded only when invoked.

### Problem Before
Previously, the top-level CLI eagerly imported all command modules:

```python
from biahub.track import track_cli
from biahub.segment import segment_cli
cli.add_command(track_cli)
cli.add_command(segment_cli)
```

This meant that running even a simple command like biahub --help triggered import of all dependencies (e.g. torch, numpy, zarr, etc.), resulting in:

```bash
time biahub --help
real    0m18.0s
```

### Solution
We now use lazy registration of commands via `click.Command` and `importlib`. Each command is registered with a short help string and a stub callback that imports and invokes the actual command only when needed.

**Key changes**

- COMMANDS is now a list of dictionaries (name, import_path, help)
- lazy_import_command() defers importing the real command until runtime
- biahub --help now runs in ~0.2s instead of 18s

### DIFF
--- a/biahub/cli/main.py
+++ b/biahub/cli/main.py
@@ -25,113 +25,127 @@ def lazy_import_command(import_path):
     return callback
 
 
-# Command list: (name, import_path, help)
 COMMANDS = [
-    (
-        "estimate_bleaching",
-        "biahub.estimate_bleaching.estimate_bleaching_cli",
-        "Estimate bleaching from raw data",
-    ),
-    (
-        "estimate_deskew",
-        "biahub.estimate_deskew.estimate_deskew_cli",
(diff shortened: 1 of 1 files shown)


ziw-liu:
What necessitates the difference between the way that this PR (lazy commands) and the example in the [docs](https://click.palletsprojects.com/en/stable/complex/#lazily-loading-subcommands)(lazy groups)?

PR #142, opened 2025-08-13
Added `napari-ome-zarr` as a dependency

srivarra:


ieivanov:
Hi Sri, can you explain the motivation here?

srivarra:
@ieivanov A user can't drag n' drop a OME Zarr store into Napari with the `biahub` conda environment unless they have `napari-ome-zarr` installed. Would this be better suited in the kitchen sink environment I'm tasked with instead of here?

PR #180, opened 2025-11-22
Generate mov cli from napari to mp4

tayllatheodoro:
# Summary

This PR introduces a new CLI command `napari-movie` that renders MP4 videos from OME-Zarr datasets using napari, with the option of just opening the data with configs. It supports:

-  Multiple inputs layered in a single viewer (image and label)
- Per-channel contrast limits and colormaps
- Z-max projection, single Z-slice, or single timepoint extraction
- Optional Ultrack tracks overlay from CSV
- Unique output filenames to avoid overwriting
- Batch processing across FOVs defined by a glob pattern
- Optional interactive mode to inspect the viewer (--view-napari)

The feature is configured via a YAML file.

# Motivation

We frequently need publication-quality videos that combine:

- Raw/virtual-stain channels,
- Segmentation labels
- Optional cell tracks

…with consistent contrast and color styling across many fields of view (FOVs). Doing this manually in Napari is time-consuming and not reproducible. This CLI provides a reproducible, config-driven workflow that integrates with our existing biahub stack.

### DIFF
--- a/tests/test_example_settings.py
+++ b/tests/test_example_settings.py
@@ -12,13 +12,13 @@ from biahub.settings import (
     DeskewSettings,
     EstimateRegistrationSettings,
     EstimateStabilizationSettings,
+    NapariMovieSettings,
     ProcessingImportFuncSettings,
     RegistrationSettings,
     SegmentationSettings,
     StabilizationSettings,
     StitchSettings,
     TrackingSettings,
-    NapariMovieSettings,
 )
 
 settings_files_dir = (Path(__file__) / "../../settings").resolve()


biahub/visualize/napari_movie.py
```diff
@@ -0,0 +1,605 @@
...
+        click.echo(f"🎬 Rendering video at {fps} FPS...")
+        writer = imageio.get_writer(output_file, fps=fps)
+        num_frames = data_dict["arr_data"][0].shape[0]
+
+        for i in tqdm(range(num_frames), desc=f"Rendering {fov}", unit="frame"):
+            viewer.dims.set_point(0, i)
+            screenshot = viewer.screenshot(canvas_only=True)
+            writer.append_data(np.array(screenshot))
+
+        writer.close()
```
srivarra: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
`writer` can be a context manager, so instead of manually closing it we can do something like

```python
with writer:
	num_frames = ...
	writer.append_data(...)
```

Therefore it doesn't need to be manually closed.

PR #191, opened 2025-12-04
Setup GitHub Pages Documentation

srivarra:
Instead of waiting for RTD access, let's just get something available. I've built the docs with [zensical](https://zensical.org/) made by the creators of [mkdocs-material](https://squidfunk.github.io/mkdocs-material/). It's modern, extensible and has a great set of default features. It's also incredibly easy to use, just being simple markdown files.

Some future work would include more tutorials, tips, etc... In addition I think a designing a Biohub theme from the brand book's guidelines would be pretty cool.

Currently I've set up just set it up for what's in the `docs/` directory, a contributing file and added a page for the CLI interfaces.

Locally you can view the documentation by first installing the doc extra `pip install .[doc]` and then running the command `zensical serve --open`.

Note, gh pages doesn't really have a good way to view the documentation on a per PR basis...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,12 @@
 # biahub
 
+[![Docs][docs-badge]][docs-url]
+
+<!-- Badges -->
+[docs-badge]: https://github.com/czbiohub-sf/biahub/actions/workflows/docs.yml/badge.svg
+<!-- URLs -->
+[docs-url]: https://czbiohub-sf.github.io/biahub
+
 Bio-image analysis hub supporting high-throughput data reconstruction on HPC clusters with [Slurm](https://slurm.schedmd.com/documentation.html) workload management.
 
 


ieivanov (approved):
Thanks Sri! Approving so you can merge at any point, now or later.

A couple of things I noted:
* once we merge and the docs build, we should add a link to the site in the biahub project description, maybe in the readme too, where it will be easy to find
* the Contributing documentation is duplicated, once at root, and again under docs/. We should keep only one copy
* It's awesome to see the documentation of the CLI. I think that's P0. As a next step, we should expose documentation on the python API. Could you trial that for the deskew methods? That can come in a second PR
* Currently the menu on the left hand side looks something like this. We should iterate on that. I'd love to see docs on the CLI and then an entry for each recon methods below (e.g. deskew, registration, stabilization, etc.). The `Correlative Confocal - QPI Imaging` and `Image Processing` sections are not as critical.

Again thanks a lot, we've been wanting this for a long time!

Image

srivarra:
@ieivanov 

> the Contributing documentation is duplicated, once at root, and again under docs/. We should keep only one copy


After looking at it again, yeah we should only have one copy for now, I think later on we should create a custom one as GH markdown is more restrictive and has less features unlike [mkdocs markdown](https://zensical.org/docs/setup/extensions/python-markdown/) and [markdown extensions](https://zensical.org/docs/setup/extensions/python-markdown-extensions/) supported in zensical.

> we should add a link to the site in the biahub project description

I can add a for this badge in this PR.


> As a next step, we should expose documentation on the python API. Could you trial that for the deskew methods? That can come in a second PR

As a first thought, we might need to do some docstring formatting, but I'll give it a try in a separate PR.


> Currently the menu on the left hand side looks something like this. We should iterate on that. I'd love to see docs on the CLI and then an entry for each recon methods below (e.g. deskew, registration, stabilization, etc.). The Correlative Confocal - QPI Imaging and Image Processing sections are not as critical.

Could you provide an overview on what types of documentation you would like? This sort of sounds like the API docs to me? Perhaps I'm confused.

PR #223, opened 2026-04-01
Add deskew overhang mean fill

### LINKED ISSUE #118: Filling overhang region with zeros creates labelfree reconstruction artifacts
`biahub deskew` with `keep_overhang=True` fills the overhang region with zeros. This is appropriate for fluorescence because the background is near zero, but this is inappropriate for labelfree data because it creates a large sharp edge.

This sharp edge creates harsh reconstruction artifacts (see left side of attached movie). To reduce this artifact, I wrote a script that fills the overhang regions with the mean of the non-overhanging regions. This fix softens the edges and nearly eliminates the reconstruction artifacts (see right side of attached). 

https://github.com/user-attachments/assets/dc77ed3b-5500-4d71-bb0c-a07da11f3de2

I propose we expose a `padding_mode="zeros" or "mean"` option in the deskew config.

tayllatheodoro:
## Summary
- **Deskew overhang fill**: Added `overhang_fill` option (`"zero"` or `"mean"`) to `DeskewSettings`. When set to `"mean"`, zero-padded overhang regions are replaced with the mean of the valid signal, reducing edge artifacts.
- **Deskew API refactor**: Separated CLI (`deskew_cli`) from the callable API (`deskew`), and renamed the core ZYX function to `deskew_zyx` for clarity.
- ~**Flat field correction**: New `biahub flat-field` CLI command that applies median-based flat field correction to selected channels, with SLURM support and per-timepoint parallelization.~

## Test plan
- [x] Verify example settings parse correctly (`test_example_settings.py`)
- [x] Run deskew with `overhang_fill: mean` and confirm overhangs are filled
- [x] Run `biahub flat-field` on a test dataset and verify corrected output


talonchandler:
Tested...this is working as expected. We may need additional iteration to find a filling strategy that minimizes artifacts near fast-changing well edges, but this is a great working start. 

I made a merge commit so that this only contains the overhang filling function.

talonchandler (approved):

PR #224, opened 2026-04-03
Example for nextflow wiring for biahub 

aofei-liu:
## Summary

Nextflow orchestration layer for biahub image processing pipelines, with modular DSL2 architecture, a comprehensive `biahub nf` CLI, and end-to-end support for the mantis v2 timelapse workflow.

### Nextflow pipelines (`nextflow/`)

- **`mantis-v2-timelapse.nf`**: Full mantis v2 pipeline — flat-field → deskew → reconstruct → virtual stain → tracking → rename channels → concatenate → stabilization → QC reporting. Gates optional steps (tracking, virtual stain) on config presence. Stabilization is available as a standalone pipeline (`stabilization.nf`) and reusable module but is not wired into the mantis v2 pipeline yet.
- **`example-flatfield-deskew-reconstruct.nf`**: Lightweight four-step example pipeline for quick runs.
- **`stabilization.nf`**: Standalone stabilization entrypoint that includes the stabilization module.

### Modular DSL2 architecture (`nextflow/modules/`)

- 10 reusable modules: `flat_field`, `deskew`, `reconstruct`, `virtual_stain`, `tracking`, `rename_channels`, `assembly` (concatenate), `stabilization`, `qc`, and `common` (shared functions).
- Each module defines processes and named subworkflows with explicit `take`/`emit` interfaces for composability.
- Shared helpers in `common.nf` (resource parsing, position fan-out).
- Dynamic per-step resource estimation via `RESOURCES:` stdout lines parsed by Nextflow.
- `nextflow.config` with `local` and `slurm` profiles, GPU queue routing, and `-resume` support.

### `biahub nf` CLI (`biahub/cli/nf.py`)

~2000 lines of single-unit-of-work CLI commands designed for Nextflow scheduling:
- **Fan-out**: `list-positions` emits position names for Nextflow to scatter.
- **Init commands**: `init-flat-field`, `init-deskew`, `init-reconstruct`, `init-deconv`, `init-stain`, `init-track`, `init-concatenate`, `init-stabilize-*` — create output stores and write per-position YAML configs. Emit `RESOURCES:` lines for dynamic resource allocation.
- **Run commands**: `run-flat-field`, `run-deskew`, `run-reconstruct`, `run-deconv`, `run-stain`, `run-track`, `run-concatenate`, `run-apply-inv-tf`, `run-stabilize-*` — execute single-position processing.
- **Utilities**: `compute-transfer-function`, `rename-channels`, `flip`, `register`, `stitch`, `init-resources`.
- Uses `uv run --project` for portable invocation without an activated venv.

### Bug fixes and improvements

- Fix `init-concatenate` failing on Zarr v3 stores; default concatenate output to OME-Zarr v0.5.
- Fix `run-track` passing raw dict instead of `MainConfig` to ultrack; monkey-patch scikit-image read-only buffer bugs.
- Fix race condition: `rename_wf` waits for VS + tracking to finish before reading reconstruct zarr.
- Fix stale viscy temp zarr on retry; unique QC summary filenames.
- Create parent directories in `model_to_yaml` before writing.
- Gate `track_wf` on virtual stain completion; wire QC stages with correct dependency ordering.
- Switch flat-field from `multiprocessing.Pool` to `ThreadPoolExecutor` to avoid OOM on large zarr datasets.

### Build and tooling

- Migrate packaging from setuptools to `uv` with `hatchling` backend.
- Replace black/isort/flake8 with `ruff`.
- Bump `iohub` to 0.3.2, `requires-python` to `>=3.12`.
- Pin `imaging-qc-pipeline` to v0.3.1; add `cupy-cuda12x` for GPU tracking.
- Skip ultrack-dependent tests when ultrack is not installed.

### QC reporting

- Integrate `imaging-qc-pipeline` into the Nextflow DAG (`qc.nf` module).
- Configurable `--static` flag via `qc_report_static` param.
- `quarto_bin` param for interactive reports on Slurm.
- `--multi-store` mode for per-stage example images with `--config` passthrough.

### Known issues (to fix in follow-up PRs)

- **`virtual_stain.py` broken import**: PR #215 removed `create_empty_hcs_zarr` from `cli/utils.py` but `virtual_stain.py` still imports it. The `test_command_help[virtual-stain]` test is skipped in this PR. The fix belongs in a separate PR that updates `virtual_stain.py` to use `iohub.ngff.utils.create_empty_plate` (or removes the legacy module entirely, since the Nextflow pipeline calls viscy directly).

## Test plan

- 1571-line test suite in `tests/test_cli/test_nf_cli.py` covering all `biahub nf` subcommands.
- Tested end-to-end on SLURM via `scripts/test-nextflow.sh`.


biahub/cli/nf.py
```diff
@@ -0,0 +1,413 @@
...
+
+    temp_position = Path(temp_zarr) / position
+    output_position = Path(output_zarr) / position
+
+    with open_ome_zarr(str(temp_position), mode="r") as src:
+        src_data = np.asarray(src[0][:])
+
+    with open_ome_zarr(str(output_position), mode="r+") as dst:
+        dst[0][:] = src_data
+
```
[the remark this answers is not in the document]
aofei-liu: realistically, probably wouldn't oom?

ieivanov:
Great first implementation, I'll look closer. A few immediate observations:
* `biahub` also has an `estimate_resources` function in `biahub.cli.utils` - which I think is identical to the waveorder implementation, no need to import from there.
* it will likely be helpful to switch the entire pipeline to `uv` first

aofei-liu:
need to drop num_processes -> num_threads in unrelated code

ieivanov:
Quick note - it better to run deskew on CPUs, it end up being faster as you can split across many more CPUs. The option to run deskew on GPU is large just for real-time visualization

pyproject.toml
```diff
@@ -48,12 +48,14 @@ dependencies = [
 [project.optional-dependencies]
 segment = ["cellpose"]
 
-track = ["ultrack>=0.7"]
+track = ["ultrack>=0.7", "cupy-cuda12x"]
 
 gui = ["napari[pyqt6]", "PyQt6>=6.5,<6.8", "napari-animation"]
 
+shard = ["tensorstore"]
```
ieivanov: @aofei-liu do you still need `tensorstore`? It seems like it's not used anywhere in the code, this may have been a merge carry-over.

pyproject.toml
```diff
@@ -48,12 +48,14 @@ dependencies = [
 [project.optional-dependencies]
 segment = ["cellpose"]
 
-track = ["ultrack>=0.7"]
+track = ["ultrack>=0.7", "cupy-cuda12x"]
```
ieivanov: Why was `cupy-cuda12x` required? This package is not available on macOS, we should use more constrained `sys_platform` or `platform_machine` specifications

tayllatheodoro: CUDA is a requirement for Ultrack, but not needed on HPC, as we do module load CUDA

pyproject.toml
aofei-liu: ahh possibly. thanks for catching

pyproject.toml
ieivanov: Did you find that cupy makes your code run faster? @edyoshikun mentioned that the latest version of ultrack achieves acceleration by using compiled binaries rather than cupy - have you tried that? Also, CUDA 13 is available on bruno from `module load cuda` but in your scripts you specifically ask for 12.8 with `module load cuda/12.8.0_570.86.10`. Is there a reason for that?

talonchandler (approved):
No concerns from me. I ran a couple example `flatfield-deskew-reconstruct` pipelines with no issues, and I looked closest at how it's calling waveorder's reconstruction calls. 

Thanks @aofei-liu @tayllatheodoro @ieivanov and co!

Reports are cool :sunglasses:
Image

ieivanov (changes requested):
@aofei-liu @tayllatheodoro thanks for putting this together — the Nextflow wiring and the end-to-end mantis v2 pipeline implementation is very exciting! During the first round of development, it was useful to build out `nf.py` and leave the existing CLI unchanged for fast debugging. Now it's time for us to refactor and clean up the code.

My main concern is that `biahub/cli/nf.py` reimplements logic that already exists in the per-step modules (`deskew.py`, `flat_field_correction.py`, `reconstruct.py`, `stabilization.py`, …). At ~2,100 lines, `nf.py` becomes a second source of truth for "how to init an output zarr" and "how to run one position" for every step. The two will drift, and we have at least one drift already — see below.

I'd like us to refactor before merging so that the per-step modules are the single source of truth, and `nf.py` (or, better, the Nextflow processes themselves) just call into them. Below I use deskew as a concrete example; the same pattern applies across the rest of `nf.py`.

> **Draft PR demonstrating the proposed refactor on the deskew step:** [#252 — refactor: consolidate deskew init/run into biahub deskew](https://github.com/czbiohub-sf/biahub/pull/252). If the approach there looks good, we can apply the same pattern for flat-field, reconstruct, stabilization, concatenate, tracking, virtual stain, register, deconvolve, and stitch.

## Concrete example: deskew duplication

Today `biahub/deskew.py::deskew` (lines 573–702) does three things in sequence:

1. Read input plate metadata.
2. Create the empty output plate with the deskewed shape and voxel size, attach `extra_metadata`.
3. Submit one `process_single_position(_czyx_fast_deskew_data, …)` job per position via submitit.

`biahub/cli/nf.py::init_deskew` (lines 140–201) reimplements step 2, and `nf.py::run_deskew` (lines 204–232) reimplements step 3 for a single position. Side-by-side:

| | `deskew.py::deskew` | `nf.py::init_deskew` + `run_deskew` |
|---|---|---|
| Load `DeskewSettings` | `yaml_to_model(config_filepath, DeskewSettings)` | duplicated in both `init_deskew` and `run_deskew` |
| Compute output shape | `get_deskewed_data_shape(...)` | duplicated in `init_deskew` |
| Create plate | `create_empty_plate(...)` with `resolve_ome_zarr_version(...)` | duplicated in `init_deskew`, **but hardcoded to `version="0.5"`** |
| Attach `extra_metadata` | via `process_single_position(..., extra_metadata=...)` per position | written by `init_deskew` directly into `zattrs` for every position, then **also** passed via `process_single_position` in `run_deskew` |
| Per-position kernel | `_czyx_fast_deskew_data` (GPU fast path) | `_czyx_deskew_data` (the slow MONAI `Affine` path) — **drift** |
| Resolve `pixel_size_um` from input scale | not done | done in `init_deskew` only — the SLURM flow doesn't benefit |

Two concrete bugs fall out of this duplication:

- **The Nextflow path runs the slow kernel.** `run_deskew` calls `_czyx_deskew_data`, but the submitit path in `deskew()` uses `_czyx_fast_deskew_data`. The fast path is the one that exists for a reason — it should be the default everywhere.
- **OME-Zarr version handling diverges.** `deskew()` honors `settings.output_ome_zarr_version` via `resolve_ome_zarr_version`; `init_deskew` hardcodes `"0.5"`.

These are the kind of drifts the duplication will keep producing.

## Suggested refactor

Make `deskew.py` accommodate the Nextflow use case directly, instead of letting `nf.py` carry its own copy. The key observation is that `iohub.ngff.utils.create_empty_plate` is already idempotent — its docstring says "If the plate exists, append positions and channels if they are not already in the plate." And `deskew()` already calls it on every run (`biahub/deskew.py:628`). So most of the machinery we need already exists; we just need to expose it cleanly.

The proposal is to manage with **just two entry points** on the existing `biahub deskew` command, instead of adding a parallel `biahub nf init-deskew` / `biahub nf run-deskew` pair:

```
biahub deskew --init -i INPUT -o OUTPUT -c CONFIG
biahub deskew        -i INPUT -o OUTPUT -c CONFIG  [--sbatch FILE] [--executor slurm|local|debug] [--no-monitor]
```

### Behavior

**`biahub deskew --init`** — initialization only. Reads plate metadata from `INPUT`, resolves `DeskewSettings` (including `pixel_size_um` from the input zarr scale if absent), calls `create_empty_plate` to create the output plate with the deskewed shape, voxel size, and `extra_metadata`. Does not process any data. This is the only step that needs to be sequenced before fan-out, because creating the plate root is the only operation that races.

**`biahub deskew`** (no `--init`) — processes one or more positions. The input is already a list of positions today (the `input_position_dirpaths` decorator uses `nargs=-1` and expands `./input.zarr/*/*/*`), so "single position" vs "whole plate" is just a list length, and we don't need to branch on it explicitly. We instead pick the execution mode via a new `--executor` flag.

### Replace `--local` with `--cluster`

The current `--local` flag (`biahub/cli/parsing.py:252`) is a boolean that flips between SLURM and a local submitit cluster. The helper `get_submitit_cluster` (`biahub/cli/utils.py:116`) already maps this to the three string values submitit's `AutoExecutor` understands; it just doesn't expose all three.

Per the submitit 1.5.4 docs (`AutoExecutor.__init__`, `submitit/auto/auto.py:48–50`):

> `cluster: str` — Forces AutoExecutor to use the given environment. Use **"local"** to run jobs locally, **"debug"** to run jobs in process.

So `AutoExecutor(cluster=...)` natively supports `"slurm"`, `"local"`, and `"debug"`. Let's expose all three:

| `--cluster` | Underlying submitit cluster | What it does |
|---|---|---|
| `slurm` (default on Bruno) | `"slurm"` | submit jobs via sbatch (current default) |
| `local` | `"local"` | run jobs as subprocesses on the local machine (the current `--local` behavior) |
| `debug` | `"debug"` | run jobs **in-process** in the foreground of the calling Python process — exactly what a Nextflow worker needs |

This is cleaner than branching on `len(input_position_dirpaths) == 1`:

- A Nextflow worker calls `biahub deskew --executor debug -i ${plate}/${position} -o ${out} -c ${cfg}` and gets a synchronous in-process run, no submitit fan-out, no monitor.
- An interactive user calls `biahub deskew --executor local ...` to test on their laptop.
- The SLURM flow is unchanged — `--executor slurm` is the default.

`get_submitit_cluster(local: bool)` becomes `get_submitit_cluster(executor: Literal["slurm", "local", "debug"])` and stays a one-liner. The `CI=true` override can either stay (it maps any executor to `"debug"`) or be removed in favor of having CI pass `--cluster debug` explicitly — your call.

For the Nextflow pipeline, the orchestration becomes:

```nextflow
process deskew_init {
    script: "biahub deskew --init -i ${plate} -o ${out} -c ${cfg}"
}
process deskew_run {
    input:  val position from positions_ch
    script: "biahub deskew --executor debug -i ${plate}/${position} -o ${out} -c ${cfg}"
}
```

`deskew_init` runs once on the plate. `deskew_run` fans out across positions, each one an in-process invocation that doesn't try to launch submitit jobs from inside a Nextflow task. No `init-deskew` / `run-deskew` in `nf.py` needed.

### What changes inside `deskew.py`

The internal refactor is small because most of this already exists:

1. Pull `pixel_size_um` resolution from the input zarr scale (currently only in `nf.py::init_deskew`, lines 163–170) into `deskew.py` so the SLURM flow benefits too.
2. Factor the plate-creation block (`deskew.py:609–637`) into a helper, e.g. `_init_output_plate(input_position_dirpaths, output_dirpath, settings)`. Both code paths call it.
3. Add `--init` as a Click flag on `deskew_cli`; when set, return after `_init_output_plate`.
4. Replace the `local()` decorator in `biahub/cli/parsing.py` with an `executor()` decorator exposing `click.Choice(["slurm", "local", "debug"])`. Update `get_submitit_cluster` to take the string directly. The existing submitit dispatch loop in `deskew()` (`biahub/deskew.py:673–693`) keeps working as-is for all three values because `AutoExecutor` handles the differences internally.
5. If the output plate doesn't exist when `--init` was not passed, init it first by calling the same code path as `--init`. `create_empty_plate` is idempotent (per the iohub docstring above), so this is safe even under SLURM fan-out.
6. Drop `_czyx_deskew_data` from the Nextflow path — every code path uses `_czyx_fast_deskew_data` (the GPU fast path that the SLURM flow already uses). If there's a real reason `nf.py` was using the slow MONAI kernel, let's call it out explicitly rather than have it be an accidental drift.

### What changes in `nf.py`

- Delete `init_deskew` and `run_deskew`.
- The `RESOURCES:` emission stays in `nf.py` as the existing generic `init-resources` command (`nf.py:48–58`) — it's already independent of any one step and doesn't need to be duplicated per step. The Nextflow process for deskew can call `biahub nf init-resources -i ${plate} -r 16 --max-num-cpus 16` separately from `biahub deskew --init` if dynamic sizing is needed.

### Apply the same pattern to the rest of `nf.py`

Every `init-*` / `run-*` pair in `nf.py` (flat-field, reconstruct, stabilization, concatenate, tracking, virtual stain, register, deconvolve, stitch, …) follows the same shape and can be collapsed the same way. After the refactor, `nf.py` should be mostly the Nextflow-only utilities — `list-positions`, `init-resources`, `clean-temp`, `clean-intermediates`, `compute-transfer-function`, `rename-channels`, `combine-transforms`, the `estimate-*` helpers — plus a few thin shims where unavoidable. I'd expect it to shrink from ~2,100 lines to a few hundred. Anything significantly bigger is a sign we're still duplicating logic the per-step modules should own.

## Why this matters

- **One implementation, one set of bugs.** The kernel drift (`_czyx_deskew_data` vs `_czyx_fast_deskew_data`) and the zarr-version drift above are exactly the failure mode duplication produces. We will not catch the next one in review either.
- **Easier to test.** `init_deskew_plate` and `deskew_single_position` are unit-testable without Nextflow or submitit in the loop.
- **`nf.py` stops being a parallel codebase.** Right now anyone touching deskew has to remember to update both files; after the refactor they only touch `deskew.py`.

Happy to pair on finishing the deskew refactor as a template for the rest. I think this is worth doing before merge because every step we leave duplicated is a future drift waiting to happen, and the refactor gets cheaper the sooner we do it (vs. after more pipelines depend on the `biahub nf` surface).

aofei-liu:
sounds good! the duplication was a result of when i was testing stuff and didn't want to touch the original cli surface so as not to block existing runs that were ongoing, but agreed that we should refactor everything.

I have a proposal - let's leave this PR (#224) up as a reference, and create smaller PRs like #252 to refactor each function separately, then grab the adapter pieces that don't fit anywhere here as a final PR. what do you think? 

also, i have some qc cli related changes in a separate development branch that merges into this one -- i might do that merge here so this branch has all of the source of truth. what do you think?

tayllatheodoro:
I agree with decoupling this pr into small prs. It will be easier to review and good for documentation

ieivanov:
Great! With #252 as an example, I'm happy to leave the implementation to you. I do like the idea of smaller PRs which are easier to review. Let me know if this will also work from the standpoint of testing the pipeline and making sure we can replicate the functionality in this branch. I know you also have other PRs into `wire-nextflow` - would these be simple to port over? Shouldn't be as hard with Claude.

I'll take another look at #252 to make sure I'm happy with that as a template for the rest of the steps. I'd appreciate if you could also run deskew with that branch to make sure nothing breaks. Thanks a bunch! I'm super excited for this, and sorry again I've been slow with the review. I appreciate that you keep pushing ahead, I'll try to keep pace with you.

pyproject.toml
edyoshikun: See https://github.com/royerlab/ultrack-td .
In my hands, this has run faster.

tayllatheodoro:
Following up on the submitit-debug concerns from @aofei-liu's [analysis](https://github.com/czbiohub-sf/biahub/blob/submitit-debug-nextflow-concerns/examples/submitit_debug_nextflow/2026-05-27-submitit-debug-nextflow-concerns.md) — I pushed an alternative executor abstraction in #256 (`JobRunner`) that resolves all three by keeping submitit out of the single-job path entirely.

The Nextflow worker call becomes `biahub deskew --executor sequential -i <one position> ...`, where `sequential` calls the function in-process (`fn(*args, **kwargs)`) — no submitit, no DebugExecutor:

1. **`pdb.post_mortem()` on exception** — gone; sequential is fail-fast, the exception propagates and the CLI exits non-zero with a real traceback.
2. **`clean_env` stripping Nextflow vars** — gone; `submitit.helpers.clean_env()` only runs in the submit path, sequential never touches the environment.
3. **shadow `submitit_logs`** — gone; sequential writes no submitit logs/job-id file, stdout/stderr go straight to the process for Nextflow to capture (no split/duplication). The DebugExecutor lazy-execution hang edge case is removed too.

Same intent as the `--cluster debug` approach, but in-process is a ackend rather than submitit's debug executor with its behavior neutralized. Happy to align the two.

aofei-liu (commented):
## Review: `copy_position_metadata` in this PR

Same function reviewed in iohub PR #419 — the following fixes were applied there and should be propagated here:

### 1. Remove label-reference copying (dangling references)
Copying `metadata.labels` writes label *names* into dest metadata, but the underlying `labels/<name>/...` array groups are not copied. Consumers that enumerate labels will find references that don't exist. Since label data needs a separate copy step anyway, skip label references entirely.

### 2. Drop `ImagesMeta` re-validation
The `ImagesMeta.model_validate` call (with hardcoded `"0.5"` fallback) was only used to read `.labels`. With label copying removed, the `from iohub.ngff.models import ImagesMeta` import and the validation block can be deleted. The custom-attrs filtering against `_OME_KEYS` is sufficient on its own.

### Suggested replacement:

```python
_OME_KEYS = {"ome", "multiscales", "omero", "labels", "version"}


def copy_position_metadata(input_zarr: Path, output_zarr: Path) -> None:
    """Copy per-position custom (non-OME) zattrs from input to output plate.

    Transfers any custom zattrs such as ``extra_metadata``. Does **not**
    copy coordinate transforms, axis definitions, or label references:
    transforms change across processing steps, and label arrays require
    a separate data-copy step.
    """
    with open_ome_zarr(str(input_zarr), mode="r") as src_plate:
        with open_ome_zarr(str(output_zarr), mode="r+") as dst_plate:
            dst_positions = {name for name, _ in dst_plate.positions()}

            for name, src_pos in src_plate.positions():
                if name not in dst_positions:
                    continue

                dst_pos = dst_plate[name]

                raw_attrs = dict(src_pos.zattrs)
                custom_attrs = {
                    k: v for k, v in raw_attrs.items() if k not in _OME_KEYS
                }

                for k, v in custom_attrs.items():
                    dst_pos.zattrs[k] = v

                logger.debug("Copied metadata for position %s", name)
```

Cross-ref: [iohub#419](https://github.com/czbiohub-sf/iohub/pull/419)

ieivanov:
Closing as implemented.

PR #246, opened 2026-05-05
feat: distributed tile-stitch

srivarra:
Signed-off-by: Sricharan Reddy Varra <sricharan.varra@biohub.org>


talonchandler:
My commit here is paired with a commit to https://github.com/mehta-lab/waveorder/pull/559.

srivarra:
@talonchandler I went through and cleaned up the code in the PR. Ran a test to make sure that it still works (it does). Do you wanna take another look before merging, or do you think it's good to merge now?

talonchandler (changes requested):
Thanks @srivarra. I think this is pretty close, but I'd like to handle some integration pieces before merging:

- I just released `waveorder==3.0.5` with your tile-stitch changes there, can you unpin the dependencies here?
- can you adhere to the the biahub cli conventions? Nextflow integration isn't needed immediately, but it is valuable to reuse the CLI decorators (e.g. `@input_position_dirpaths`) like the rest of biahub. Using the decorators should address most of the convention changes: specifically, inputs should be lists of positions, --timepoint/--channel should be at the CLI level.
- can you keep the settings in `biahub/settings/.py` like the others?
- please document that we're not including this in the nextflow pipeline in this pass.

srivarra:
@talonchandler Okay, I believe I've handled all the feedback. For documenting that it won't be apart of the nextflow pipeline, I've added that to the cli description, let me know if that's a better place.

> inputs should be lists of positions, --timepoint/--channel should be at the CLI level.

I don't actually see this in any of the other cli's here, everything else seems to use the config file for those. So I've left them in the in the config file.

ieivanov:
Let me know when you guys think this PR is ready to merge. For now I'll convert it to draft.

PR #249, opened 2026-05-15
Add cellpose segmentation and waveorder focus finding to tracking

tayllatheodoro:
## Summary
- Add cellpose segmentation mode to tracking as alternative to foreground+contour
- Add waveorder-based focus finding (`FocusConfig`) for automatic z-slice selection with `z_window`
- Add `CellposeConfig` and `FocusConfig` settings, wired through CLI and SLURM submission

## Test plan
- [x] Run `biahub nf run-track` with cellpose config on single FOV
- [x] Verify waveorder focus finding selects correct z range
- [x] Run full pipeline on all positions

### DIFF
--- a/biahub/track.py
+++ b/biahub/track.py
@@ -18,12 +18,11 @@ import pandas as pd
 import submitit
 import toml
 
-
+from cellpose import models as cp_models
 from iohub import open_ome_zarr
 from iohub.ngff.utils import create_empty_plate
 from numpy.typing import ArrayLike
-from cellpose import models as cp_models
-
+from tqdm import tqdm
 
 from biahub.cli.parsing import (
     config_filepath,
@@ -40,8 +39,12 @@ from biahub.cli.utils import (
     update_model,
     yaml_to_model,
 )
-from biahub.settings import CellposeConfig, FocusConfig, ProcessingInputChannel, TrackingSettings
-from tqdm import tqdm
+from biahub.settings import (
+    CellposeConfig,
+    FocusConfig,
+    ProcessingInputChannel,
+    TrackingSettings,
+)
 
 # Lazy imports for ultrack - imported only when needed in specific functions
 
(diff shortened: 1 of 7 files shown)


biahub/track.py
```diff
@@ -342,82 +349,177 @@ def resolve_z_slice(z_range: tuple[int, int], z_shape: int) -> tuple[slice, int]
     return z_slices, Z
 
 
+def find_focus_per_timepoint(
```
edyoshikun, line 352: IMO we should just reuse the utility in viscy and store the metrics in the `ngff.Position`. We use this find focus multiple times. https://github.com/mehta-lab/VisCy/blob/modular-viscy-staging/applications/qc/configs/focus.yml

https://github.com/mehta-lab/VisCy/tree/modular-viscy-staging/applications/qc

edyoshikun: it helps if this is configurable so we can decide if we want to use Phase or other channelss. for the nueromast if you only label the hair cells you will eventually want to do make sure the focal plane is right at those cells.

edyoshikun: the one in viscy is already using cuda so it takes ~10-15 min per mantis dataset

biahub/track.py
```diff
@@ -342,82 +349,177 @@ def resolve_z_slice(z_range: tuple[int, int], z_shape: int) -> tuple[slice, int]
...
+    """Run cellpose on each 2D frame of a (T, Y, X) array.
+
+    Returns an integer label array of shape (T, Y, X).
+    """
+    model = cp_models.CellposeModel(model_type=model_type, gpu=gpu)
+
+    T = images.shape[0]
+    labels = np.zeros_like(images, dtype=np.int32)
+    for t in tqdm(range(T), desc="Cellpose segmentation"):
+        mask, _, _ = model.eval(
```
edyoshikun, line 473: thiis is slow. I think we can exploit and add more images in the batch dimensions, so you process multiuple at the same time instead of one timepoint at the time

edyoshikun: also take a look at OPS team's optimization that mark did https://github.com/royerlab/ops_process/blob/78aba28ab650421ce73dfc26c5edde669b0ab362/ops_analysis/processes/segment.py#L399

biahub/track.py
```diff
@@ -342,82 +349,177 @@ def resolve_z_slice(z_range: tuple[int, int], z_shape: int) -> tuple[slice, int]
...
+    cellprob_threshold: float = 0.0,
+    flow_threshold: float = 0.4,
+    gpu: bool = True,
+    min_size: int = 500,
+) -> np.ndarray:
+    """Run cellpose on each 2D frame of a (T, Y, X) array.
+
+    Returns an integer label array of shape (T, Y, X).
+    """
+    model = cp_models.CellposeModel(model_type=model_type, gpu=gpu)
```
edyoshikun, line 468: we shouldn't instantiate the model multiple times because it is slow. see my comment below.

biahub/track.py
```diff
@@ -701,6 +803,82 @@ def data_preprocessing(
...
+        Cellpose model parameters including ``input_channel``.
+    """
+    fov = "/".join(position_key)
+
+    data_dict = load_data(
+        position_key=position_key,
+        input_images=input_images,
+        z_slices=z_slices,
+    )
+    data_dict = run_preprocessing_pipeline(data_dict, input_images)
```
edyoshikun, line 839: can we also apply CLAHE before segmetnation?

settings/nextflow_templates/mantis_v2_track.yml
```diff
@@ -1,8 +1,28 @@
...
+  lambda_ill: 0.500
+  pixel_size: 0.1133
+  z_window: 40
+
+# Segmentation method: "foreground_contour" or "cellpose"
+segmentation_method: cellpose
+
+cellpose_config:
+  model_type: nuclei
+  diameter: 80
```
edyoshikun, line 18: I think cellpose-sam deprecated some of these parameters so just double check.

edyoshikun: can we also do preprocessing here like we had with the foreground computation?

ieivanov:
@tayllatheodoro is this PR superseded by #281? Can we close this PR?

PR #252, opened 2026-05-27
refactor: consolidate deskew init/run into biahub deskew

ieivanov:
## Summary

Demonstrates the refactor proposed in #224 review on a single step (deskew). Removes the duplicated `nf init-deskew` / `nf run-deskew` commands by extending the existing `biahub deskew` CLI with two flag-driven modes, so `biahub/cli/nf.py` and `biahub/deskew.py` stop being two parallel implementations of the same logic.

If this lands well, the same pattern can be applied to flat-field, reconstruct, stabilization, concatenate, tracking, virtual stain, register, deconvolve, and stitch in follow-up PRs — the duplication audit is in [`plans/pr224-review.md`](../tree/refactor-deskew-init-executor/plans/pr224-review.md).

### CLI surface

\`\`\`
biahub deskew --init -i INPUT -o OUTPUT -c CONFIG
biahub deskew        -i INPUT -o OUTPUT -c CONFIG [--cluster slurm|local|debug] [--sbatch FILE] [-m]
\`\`\`

- **\`--init\`**: create (or extend) the output plate and exit. Emits a \`RESOURCES:\` line for Nextflow to parse. Resolves \`pixel_size_um\` from the input zarr scale if not in the config — both flows benefit, not just Nextflow.
- **\`--cluster {slurm,local,debug}\`**: replaces \`--local\`. Maps directly to submitit \`AutoExecutor(cluster=…)\` — \`debug\` runs jobs in-process in the foreground, which is what every Nextflow worker now uses.
- \`create_empty_plate\` is idempotent (per iohub's docstring), so the plate-init code path is safe to call on every invocation; no explicit \`--init\` is required outside the Nextflow flow.

### What changed

- **\`biahub/deskew.py\`**: factored \`_load_deskew_settings\` and \`_init_output_plate\` helpers out of \`deskew()\`. \`deskew_cli\` gets \`--init\` and \`--cluster\`. Dropped \`--local\` from this command only; the \`local()\` decorator stays in \`parsing.py\` for the other CLIs.
- **\`biahub/cli/parsing.py\`**: added \`cluster()\` decorator with \`click.Choice([\"slurm\", \"local\", \"debug\"])\`.
- **\`biahub/cli/utils.py\`**: \`get_submitit_cluster\` now accepts \`cluster: str\` in addition to the legacy \`local: bool\` (backward-compat, other CLIs unaffected).
- **\`biahub/cli/nf.py\`**: deleted \`init-deskew\` and \`run-deskew\` (~95 lines).
- **\`nextflow/modules/deskew.nf\`**: \`init_deskew\` process now calls \`biahub deskew --init\`; \`run_deskew\` calls \`biahub deskew --cluster debug -i \${plate}/\${position}\`. Drops the intermediate \`deskew_resolved.yml\` file.
- **Bonus drift fix**: \`nf.run-deskew\` was using the slow MONAI \`_czyx_deskew_data\` kernel while the SLURM path uses \`_czyx_fast_deskew_data\`. Both paths now use the fast kernel.

### Test plan

- [x] \`pytest tests/test_cli/\` — 93 pass, 2 skipped (existing skips).
- [x] New tests for \`--init\`, \`--cluster local\`, \`--cluster debug\` single-position in \`tests/test_cli/test_deskew_cli.py\`.
- [ ] End-to-end Nextflow run on Bruno (not yet tested in this PR).


aofei-liu:
small concerns here: https://github.com/czbiohub-sf/biahub/blob/submitit-debug-nextflow-concerns/examples/submitit_debug_nextflow/2026-05-27-submitit-debug-nextflow-concerns.md  

tl;dr, some log pollution/messiness, potential to hang on weird edge case (interactive run, i doubt we would do that though), and the environment clearing on retry logic in nextflow could potentially be problematic if we rely on any environment variables to persist. not technically blocking, so ok to go ahead with the --debug tag, but it would potentially make error triaging more annoying.

how to reproduce logs, etc with the toy example is documented here: https://github.com/czbiohub-sf/biahub/blob/submitit-debug-nextflow-concerns/examples/submitit_debug_nextflow/README.md

ieivanov:
Thanks for looking into this @aofei-liu, super useful. Overall, I agree with you that the three concerns are not a blocker. Here is my response for each:

`pdb.post_mortem()` obfuscates errors on exception - makes error handling a bit messy, but that's largely non-issue for us. When running the CLI thru Nextflow, we can largely assume that all calls work correctly. Debugging of individual CLI calls should happen outside of Nextflow

`os.environ clear/restore can strip Nextflow-injected variables` - true for any env vars set *after* job init and *before* job execution. Quite unlikely edge case, I agree with Claude that's not an issue

`Shadow log directory splits logging away from Nextflow`- this is most interesting. If I understand correct, we can duplicate logging - both in Nextflow and in the submitit_logs directory. That's OK with me. I do want to confirm that Nextflows captures the execution logs (e.g. `Processing...` or `Finished writing...`). Let's look carefully at the Nextflow logs with this workflow.

I'm overall happy with the deskew structure in this branch. What do you think of my suggestion [on Slack](https://biohub.slack.com/archives/C09E5M6BJ3U/p1779926956661869?thread_ts=1779829990.027729&cid=C09E5M6BJ3U) to create a minimal nextflow workflow - for now deskew only, that we slowly build out, e.g. adding flatfielding and recon as next steps? Could you start a draft PR like that and we'll build out from there?

aofei-liu:
yeah logging will be duplicated/have some random pdb calls in them. I don't know what the logging behavior will be like if e.g. the internal executor loop fails really fast before the nextflow executor sets up the logging overhead - we might end up with situations where the logs exist in the submitit logs but not nextflow. I think i saw something like this while testing with the toy path, but on rerun it captured the logs, so i'm not sure if it was a one-off thing.

re- workflow: I think we were saying the same thing - i'll add the nextflow subworkflow to this branch corresponding to deskew, which i will run and test on the test data, and each additional function will be ported as a subworkflow the same way in a separate PR then i'll have a PR for the mantis-v2 pipeline which will import all the subworkflows, as is consistent with the current pattern. sound good?

ieivanov:
OK, let's keep an eye out on the logs as we crunch thru some datasets.

re- workflow: please draft a PR and I'll take a look; I think we'll be able to get thru these quickly after the first one. Thanks again

ieivanov:
Closing in favor of #255

PR #256, opened 2026-05-30
Add JobRunner executor abstraction and nest deskew settings

tayllatheodoro:
## Summary

Introduces a single, backend-agnostic way to run a function across a batch of jobs, and removes the per-CLI submitit/SLURM boilerplate that had started to drift between modules. Migrates `deskew`, `beads`, and `stabilize` onto it as the first consumers.

This is an alternative take on the executor direction explored in #252: rather than threading a `--cluster {slurm,local,debug}` flag through each CLI's hand-rolled submitit block, it factors the submission into one reusable abstraction so every CLI shares the same code path.

### `JobRunner(fn, items, fn_args)` — `biahub/cli/job_runner.py`
A job is `fn(*item, **fn_args)`. `fn` and `fn_args` are constant across the batch; `items` is the fan-out axis (a tuple is unpacked, a scalar is one positional arg). Bind constant leading positionals with `functools.partial`. Backends:

- **`sequential`** — in-process, serial, fail-fast. No SLURM, no submitit. For debugging and orchestrators that already own the fan-out (e.g. a Nextflow worker on one position).
- **`local`** — parallel subprocesses via submitit.
- **`slurm`** — parallel cluster jobs via submitit.

The kernel and its call signature are **identical** across all backends — only the wrapping differs — so the "two parallel implementations drift apart" bug (which #252 also caught in the deskew kernel) becomes structurally impossible. In CI, submitit backends transparently downgrade to `sequential`.

### `resolve_slurm_args(...)` — `biahub/cli/slurm.py`
`slurm.py` is slimmed to the submitit engine (`SlurmExecutor`, which now derives its log dir from `output_dirpath`). `resolve_slurm_args(base, shape, *, ram_multiplier, max_num_cpus)` sizes `slurm_cpus_per_task`/`slurm_mem_per_cpu` to the data over a base dict — the pattern repeated in ~15 CLIs. It takes a plain dict (not a `SlurmSettings`), so it works whether the base comes from a settings model or a literal, and preserves extra keys like `slurm_use_srun`.

### Nested `DeskewParams` — `biahub/settings.py`
`DeskewSettings` now nests a `DeskewParams` sub-model that matches the `fast_deskew_zyx` kernel signature exactly, so `deskew_params.model_dump()` splats straight into the kernel call (no hand-maintained settings→kernel mapping). Top level keeps `pixel_size_um`/`scan_step_um`, `slurm_settings`, and the output version; `px_to_scan_ratio` derivation moves to a `model_validator`.

> ⚠️ **Breaking config change:** existing deskew YAMLs must move the kernel fields under `deskew_params:` (see `settings/example_deskew_settings.yml`).

### CLI surface
`biahub deskew` replaces `--local` with `--executor {sequential,local,slurm}`.

## Migrated CLIs
- **deskew** — nested config → `_init_output_plate` → `_resolve_settings`/`resolve_slurm_args` → `JobRunner`.
- **beads** — per-timepoint estimation through `JobRunner`; SLURM args via `resolve_slurm_args`.
- **stabilize** — SLURM args via `resolve_slurm_args`.

The remaining `estimate_resources`+slurm-dict CLIs (deconvolve, concatenate, flat-field, process_data, track, pyramid, …) can adopt `resolve_slurm_args` the same way in follow-ups. `segment.py` is the one exception — it post-processes the cpu count, so it needs more than a drop-in.

## Test plan
- [x] `ruff check` + `ruff format` clean across all touched files.
- [x] `pytest tests/test_cli/test_deskew_cli.py` (incl. new `--executor sequential` case), `test_example_settings.py`, `test_cli/test_register_cli.py`, `test_cli/test_stabilization_cli.py` — all pass.
- [ ] End-to-end run on real data (`sequential` single position + `slurm` fan-out) — in progress.

## Commits
1. `refactor(slurm)` — JobRunner + resolve_slurm_args + `--executor`
2. `refactor(settings)` — nest DeskewParams
3. `refactor` — migrate deskew, beads, stabilize

### DIFF
--- a/biahub/cli/job_runner.py
+++ b/biahub/cli/job_runner.py
@@ -1,0 +1,177 @@
+import os
+
+from collections.abc import Callable, Iterable
+from pathlib import Path
+
+import click
+
+from biahub.cli.slurm import SlurmExecutor
+
+# The execution backends ``JobRunner`` dispatches over. The kernel (``fn``) and its
+# per-job call signature are identical across all three -- only the wrapping differs
+# (direct call vs. submitit local subprocess vs. submitit SLURM job).
+EXECUTORS = ("sequential", "local", "slurm")
+
+
+class JobRunner:
+    """
+    Run one ``fn`` across a batch of jobs on a chosen backend.
+
+    A job is ``fn(*item, **fn_args)``: ``fn`` and ``fn_args`` are constant across the
+    whole batch, and ``items`` is the fan-out axis -- one job per element. Each element
+    supplies that job's *varying* positional args (a tuple is unpacked; a scalar is
+    passed as a single argument). Bind any constant *leading* positional args into
+    ``fn`` with :func:`functools.partial`::
+
+        # deskew: kernel + kwargs constant, (input, output) paths vary
+        JobRunner(
+            partial(process_single_position, deskew_kernel),
+            zip(inputs, outputs, strict=True),
+            kernel_kwargs,
+            executor="slurm",
+            output_dirpath=output_dirpath,
+            slurm_args=slurm_args,
+        ).execute(monitor_paths=inputs)
+
+        # estimate: everything constant except the timepoint index
+        JobRunner(
+            estimate_tzyx, range(T), shared_kwargs, executor="slurm", output_dirpath=...
+        ).execute(finalize="wait")
+
+    ``fn`` and its call signature are *identical* across all backends; only the
+    wrapping differs. No backend may substitute a different kernel or call shape, which
+    makes the "two parallel implementations drift apart" bug structurally impossible.
+
+    Backends (``executor``):
+
+    - ``"sequential"`` -- in-process, one job at a time, fail-fast. No SLURM, no
+      submitit, no monitoring. For debugging and for orchestrators that already own
+      the fan-out (e.g. a Nextflow worker processing a single position).
+    - ``"local"`` -- parallel subprocesses on this machine via submitit.
+    - ``"slurm"`` -- parallel jobs on a SLURM cluster via submitit.
+
+    Parameters
+    ----------
+    fn : Callable
+        The callable run for every job. Constant across the batch.
+    items : Iterable
+        The fan-out axis: one job per element. Each element is that job's varying
+        positional args -- a tuple is unpacked into ``fn``, anything else is passed
+        as a single positional argument.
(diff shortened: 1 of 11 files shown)


ieivanov:
Thanks @tayllatheodoro. I do like the idea of `--executor {sequential,local,slurm}`. As for the `JobRunner`, I think it add more complexity rather than simplify it. As Shalin mentioned as our last meeting, we should focus on algorithms rather than pipeline engineering. For now I'm inclined to do the minimal work that's needed.

tayllatheodoro:
Can we bring this discussion to the group? this is a change that we discuss many times and would reduce the code duplication in biahub and improve readbility and implementation of new pipelines using slurm

ieivanov:
Ok, please bring it up for discussion at group meeting tomorrow

ieivanov:
@talonchandler, Taylla meant to bring up this PR in group meeting, but we didn't get around to it. Can you please take a look? I think it brings in more complexity than is worth it. If you agree we'll close it. Some related discussion in https://github.com/czbiohub-sf/biahub/pull/252#issuecomment-4567815999 and https://biohub.slack.com/archives/CA1U02TC2/p1780434674739679?thread_ts=1779984024.978019&cid=CA1U02TC2. Thanks!

talonchandler:
Hi @tayllatheodoro (cc @ieivanov). I've read this PR and #252 and my understanding is that both PRs propose an abstraction that ensure `biahub`'s CLIs stays consistent, and only this PR's proposal uses a `JobRunner` class which keeps things marginally tidier (I'm estimating this PR's approach saves ~20 lines per CLI). Additionally, I think this PR proposes a backwards-incompatible change to the deskew settings, but AFAICT this is not a necessary change. @tayllatheodoro, is this a fair summary?

If my understanding is correct, then I'm with @ieivanov. I would prioritize the most challenging parts of this migration (I don't have the full story, but I think getting eyes on the migrated code so that we can get everything merged to main ASAP is the current bottleneck). After we have a nextflow pipeline on main, then I think the pipeline owner can have their agent eat these kinds of refactors for breakfast.

ieivanov:
Closing this PR after our discussion this morning. @tayllatheodoro will open an issue to track the outstanding problem - we would ensure that all biahub CLIs follow the same signature, possibly thru a Claude skill.

PR #257, opened 2026-06-01
refactor: flat-field --init/--cluster + Nextflow subworkflow

aofei-liu:
## Summary

- Refactor `biahub flat-field` CLI with `--init` (create output store, emit `RESOURCES:`) and `--cluster {slurm,local,debug}` (replaces `--local`) flags
- Factor `_init_output_plate()` and `_resolve_target_indices()` helpers from the monolithic `flat_field()` function
- Add Nextflow `flat_field.nf` subworkflow: `init_flat_field → parse_resources → fan-out run_flat_field × N positions`

This is PR 1 in the [refactor plan](https://github.com/czbiohub-sf/biahub/pull/224) sequence, following the deskew pattern established in PR #255.

**Depends on:** PR #255 (shared infrastructure: `cluster()`/`init_only()` decorators, `common.nf` helpers)

## `--cluster debug` on the Nextflow path

Nextflow handles per-position fan-out and resource scheduling, so the CLI must run its work in-process via submitit's `DebugExecutor` rather than submitting its own SLURM jobs. See `flat_field.nf` header comment and `examples/submitit_debug_nextflow/2026-05-27-submitit-debug-nextflow-concerns.md`.

## Test plan

- [x] 3/3 pytest unit tests pass (init_only, debug single position, local with sbatch override)
- [x] CLI integration on Bruno: `--init` emits `RESOURCES:15 150`, creates 4 positions; `--cluster debug` processes all 4 positions (BF - Oblique corrected, 2 fluorescence channels copied as-is)
- [x] Nextflow subworkflow on Bruno via Slurm: `list_positions → init_flat_field (1/1 ✔) → run_flat_field (4/4 ✔)`, 2m 51s

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #255, opened 2026-05-29
refactor: deskew --init/--cluster + Nextflow subworkflow

aofei-liu:
## Summary

- Refactor `biahub deskew` CLI with `--init` (create output store, emit `RESOURCES:`) and `--cluster {slurm,local,debug}` (replaces `--local`) flags
- Add `fast_deskew_zyx`: 2D grid_sample exploiting axis decomposition, GPU-side averaging and overhang fill
- Add shared CLI infrastructure: `cluster()`/`init_only()` decorators, `get_submitit_cluster()`, `resolve_ome_zarr_version()`
- Add Nextflow `deskew.nf` subworkflow: `init_deskew → parse_resources → fan-out run_deskew × N positions`
- Add `common.nf` shared helpers: `biahub_cmd`, `parse_resources`, `dataset_name`, `slurm_logs`

This is the reference implementation (PR 0) for breaking PR #224 into smaller, independently mergeable PRs. Each subsequent step (flat-field, reconstruct, etc.) follows the same pattern.

## `--cluster debug` on the Nextflow path

Nextflow handles per-position fan-out and resource scheduling, so the CLI must run its work in-process via submitit's `DebugExecutor` rather than submitting its own SLURM jobs. See `deskew.nf` header comment and `examples/submitit_debug_nextflow/2026-05-27-submitit-debug-nextflow-concerns.md`.

## Test plan

- [x] 6/6 pytest unit tests pass (init_only, debug single position, averaging, overhang error)
- [x] CLI integration on Bruno: `--init` resolves pixel_size_um from zarr, emits `RESOURCES:8 472`; `--cluster debug` processes 4 positions, output shape matches reference
- [x] Nextflow subworkflow on Bruno: `list_positions → init_deskew (1/1 ✔) → run_deskew (4/4 ✔)`, 2m 37s

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #258, opened 2026-06-01
refactor: reconstruct --init/--compute-tf/--cluster with Nextflow subworkflow

aofei-liu:
## Summary

- Add `--init` and `--cluster` flags to `biahub apply-inv-tf`:
  - `--init`: create output plate, resolve pixel sizes from zarr metadata into `reconstruct_resolved.yml`, emit `RESOURCES:` and `TF_RESOURCES:`
  - `--cluster debug`: apply inverse TF per-position in-process (Nextflow fan-out)
- Add `nextflow/modules/reconstruct.nf` with 3-phase wiring: `init_apply_inv_tf → compute_transfer_function (TF_RESOURCES) → fan-out run_apply_inv_tf (RESOURCES) × N positions`
- Unit tests for `--init` and `--cluster debug` modes

### Architecture note

The Nextflow `reconstruct.nf` module calls the decomposed steps directly:
- `biahub apply-inv-tf --init` / `--cluster debug` (from `apply_inverse_transfer_function.py`)
- `biahub compute-tf` (from `compute_transfer_function.py`, unchanged)

**`reconstruct.py` is not touched by the Nextflow module** and remains as the original thin convenience wrapper (`compute-tf` → `apply-inv-tf` in one command). It can be safely deleted if desired — nothing depends on it for the Nextflow path.

### Config resolution behavior

`--init` always resolves pixel sizes from input zarr metadata and writes `reconstruct_resolved.yml`. This is **non-destructive**: existing `yx_pixel_size` / `z_pixel_size` values in the config are preserved; only missing or null values are filled in.

When resolution fills in a value, a `WARNING`-level log message is emitted so users know the config was incomplete and values came from the input data. When all values are already present, a message confirms no resolution was needed.

**Rationale for auto-resolution:** minimizes continuous workflow breaks due to config errors, especially when the correct pixel sizes are already available in the zarr metadata. The deskew step writes voxel sizes into the output zarr scale, so downstream reconstruction configs should not need to redundantly specify them.

### Cascade chain

This PR is part of the #224 refactor cascade:
- **#255** (deskew) ← base
- **#257** (flat-field) ← depends on #255
- **This PR** (reconstruct) ← depends on #257

### Gate check (from plan doc)

> Unit-of-work: the TF computation is a one-shot, cross-position job, not `process_single_position`. Confirm the run/apply-inv-tf path is the only per-position unit the Nextflow `--cluster debug` process wraps, and that TF compute is handled as its own process.

Confirmed. `--cluster debug` wraps only `apply_inverse_transfer_function_single_position` (per-position). TF compute is the standalone `biahub compute-tf` command (separate NF process, one-shot from first position).

### Also included (from upstream cascade)

- `read_plate_metadata` and `copy_position_metadata` helpers in `biahub/cli/utils.py` (from #255)
- All `_init_output_plate` functions call `copy_position_metadata` to preserve per-position OME metadata

## Test plan

- [x] `pytest tests/test_cli/test_reconstruct_cli.py` — 2/2 pass (init, debug single position with full init→compute-tf→apply pipeline)
- [x] `pytest tests/test_cli/test_deskew_cli.py tests/test_cli/test_flat_field_cli.py` — 9/9 pass (no regressions)
- [x] Tier 2: CLI integration on real mantis v2 data — 4 positions reconstructed, Pearson r=1.0 vs reference
- [x] Tier 3: `bash refactor-tests/scripts/test-reconstruct-refactor.sh nf` (Nextflow subworkflow)


ieivanov:
@aofei-liu I agree that it would be nice to inherit pixel size metadata from the input zarr stores, but @talonchandler and I agree that this is a bit trickier than it seems, so we decided to leave that feature for a future PR: https://github.com/czbiohub-sf/biahub/issues/270. For now we'll require that the config files have the correct pixel size metadata.

ieivanov (approved):

PR #259, opened 2026-06-01
refactor: virtual-stain --init/--copy with Nextflow subworkflow

aofei-liu:
## Summary

- Add `--init` and `--copy` modes to `biahub virtual-stain` CLI, absorbing `init-virtual-stain` and `copy-virtual-stain` from `nf.py`
- `--init` creates the output plate with `<channel>_prediction` channels from predict config, copies per-position OME metadata, emits `RESOURCES:` line
- `--copy` copies viscy prediction data from a temp per-position zarr into the output plate, then removes the temp zarr
- Predict config uses the shared `-c` option (replaces old `--predict-config-filepath`)
- `-i` is now optional (not needed for `--copy` mode)
- New `nextflow/modules/virtual_stain.nf` subworkflow: init → preprocess (viscy) → fan-out (predict + copy) × N positions
- VisCy preprocess/predict remain as external commands called directly from Nextflow processes — only init and copy are biahub CLI commands

**Chained PR** — base is `refactor-reconstruct` (#258).

## Test plan

- [x] Tier 1: 2/2 pytest tests pass (`test_virtual_stain_cli_init_only`, `test_virtual_stain_cli_copy`)
- [x] Tier 2: `--init` against real data — channels/shape/scale match reference exactly, emitted `RESOURCES:10 120`
- [x] No regressions: all 11 existing CLI tests (deskew, flat-field, reconstruct) still pass
- [x] Tier 3: Nextflow subworkflow end-to-end on GPU nodes — `list_positions (1/1) → init (1/1) → preprocess (1/1) → run_virtual_stain (4/4)` completed in 20m 56s, zero failures

### Tier 3 output verification
- 4 positions, channels=`['nuclei_prediction', 'membrane_prediction']`, shape=`(5, 2, 86, 1600, 1370)`
- Scale matches reference: `[1.0, 1.0, 0.17, 0.1133, 0.1133]`
- Pearson r vs reference ≈ 0.80/0.71 — expected divergence because the reference `3-virtual-stain` was produced from a reconstruct output with a duplicate-channel bug (see below)

### Data fixes (outside this PR)
- **Reference `2-reconstruct`**: had duplicate identical channels (`Phase3D_recon_recon`, `Phase3D_recon`). Fixed to single `Phase3D` channel. Old store backed up.
- **`predict.yml`**: `ckpt_path` pointed to a stale path; fixed to include the `4gpu_bf16_bs16_to_ep7/` subdirectory.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #260, opened 2026-06-01
refactor: rename-channels Nextflow subworkflow (track superseded by #281)

aofei-liu:
## Summary

- **Track:** add `--init` and `--cluster` (slurm/local/debug) to the unified `biahub track` CLI, plus `--input-images-path` for Nextflow to override config input paths. Factor `_init_output_plate` helper following the established pattern from deskew/flat-field/reconstruct.
- **Rename-channels:** extract from `nf.py` into standalone `biahub rename-channels` command (metadata-only, no data copy).
- **Nextflow modules:** `tracking.nf` calls `biahub track --init` / `biahub track --cluster debug`; `rename_channels.nf` calls `biahub rename-channels`.
- **Monkey-patch:** port `_patch_ultrack_readonly_buffer()` from PR #224 (wire-nextflow). scikit-image 0.26's Cython `_map_array` rejects read-only memoryviews that numpy 2.x produces from pandas columns and `np.asarray` results. The patch forces writable copies before passing to Cython. Remove when scikit-image >= 0.27 is available.

Continues the PR #224 refactor cascade: #255 (deskew) → #257 (flat-field) → #258 (reconstruct) → #259 virtual-stain → **this PR (tracking + rename-channels)**.

## Test results

| Tier | Test | Result |
|------|------|--------|
| 1 | `test_track_cli_init_only` | ✔ |
| 1 | `test_track_cli_debug_single_position` | ✔ |
| 1 | 3× rename-channels tests | ✔ |
| 2 | `--init` on real data | ✔ 4 positions, shape=(5,1,1,1600,1370), RESOURCES:5 30 |
| 2 | `--cluster debug` single position | ✔ 14 tracked objects |
| 3 | NF: `list_positions → init_track (1/1) → run_track (4/4)` | ✔ ~30-40s/position |

Label counts differ from reference (14/13/15/15 vs 19/18/18/15) — expected, since this PR uses corrected virtual-stain data from the refactored pipeline rather than the original buggy duplicate-channel reconstruct.

## Note on the scikit-image monkey-patch

Without `_patch_ultrack_readonly_buffer()`, ultrack's CBC solver fails with `ValueError: buffer source array is read-only` on every tracking run. This is the same fix already present in PR #224's `nf.py` — ported here into `track.py` so the unified CLI carries the workaround. The root cause is scikit-image #6378 (Cython `_map_array` rejects read-only memoryviews); it is fixed upstream in scikit-image >= 0.27.

## Test plan

- [x] Tier 1: pytest unit tests (init, debug, rename-channels)
- [x] Tier 2: CLI integration on real mantis v2 data
- [x] Tier 3: Nextflow subworkflow end-to-end (GPU, 4 positions)
- [x] Verify existing deskew/flat-field/reconstruct tests still pass


aofei-liu:
Worth exploring: @edyoshikun [suggested](https://github.com/czbiohub-sf/biahub/pull/224#discussion_r2106258889) trying [`ultrack-td`](https://github.com/royerlab/ultrack-td) — a C++ rewrite of ultrack's core segmentation that runs significantly faster. It has a v1 compatibility shim (`ultrack_td.v1.track`) that accepts `MainConfig`/foreground/contours, though the export path (labels, tracks DataFrame) would need reworking to use the `tracksdata` graph API. Could be a follow-up after this lands.

ieivanov:
The tracking work from this PR is superseded by #281, which was preferred — tracking was split out so it can be tested and merged on its own.

`rename-channels` was **not** moved to #281 and still lives here; it may be merged later (needs further discussion). Keeping this open as a draft for that piece.

ieivanov:
Channel renaming is metadata operation, doesn't need Nextflow. Currently happens with python script, we are working to name channels correctly at the start such that renaming is not needed.

PR #262, opened 2026-06-01
refactor: QC generate-report-spec + Nextflow QC subworkflow

aofei-liu:
## Summary

- Move `generate-report-spec` from `nf.py` into `biahub/qc.py` as standalone `biahub generate-report-spec` command
- Bring `qc.nf` and `qc_processes.nf` Nextflow modules from wire-nextflow (#224)
- Update NF `generate_report_spec` process to call unified CLI (`biahub generate-report-spec`) instead of `biahub nf generate-report-spec`

QC wraps the external `imaging-qc-pipeline` CLI, so it doesn't follow the `--init`/`--cluster` pattern used by other steps. Only report-spec generation is biahub-owned orchestration logic.

**Chained PR** — base is `refactor-assembly` (PR #261).

## Test plan

- [x] 6/6 pytest unit tests pass (`test_qc_cli.py`)
- [x] Tier 2: `generate-report-spec` produces correct YAML from 3 real zarr paths
- [x] Tier 3: Nextflow QC subworkflow (deferred — requires `imaging-qc-pipeline` + QC stage configs)
- [x] `ruff check` + `ruff format` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #263, opened 2026-06-01
refactor: end-to-end mantis-v2-timelapse.nf pipeline and nf CLI utilities

aofei-liu:
## Status update

The base of this PR (`refactor-qc`, #262) has been **closed** — the QC work now targets `main` directly via **#298** (imaging-qc v0.4 QC subworkflow + templates, purely additive, decoupled from the refactor-assembly chain).

**This branch is not yet rebased onto `main`, and doing so is more than a rebase.** `refactor-e2e` was branched before the QC reconcile (#295/#297) and its `mantis-v2.nf` is a complete multi-entry pipeline (`full` / `from_*`, track-before-assemble, integrated QC via `run_qc_wf`) that `include`s modules **not present on `main`**:

- `nextflow/modules/rename_channels.nf` + `biahub/rename_channels.py`
- `assemble_wf_mantisv2` (main has `assemble_wf`)
- `clean_intermediates` (main's `common.nf` lacks it)

It also diverges from main's `mantis-v2.nf` on fundamental wiring (assemble-before-track vs track-before-assemble; `virtual_stain_config` vs `predict_config`).

### Path to main
1. Land #298 (QC → main). ✅ open
2. Port the remaining unmerged modules to main (rename-channels CLI + `.nf`, `assemble_wf_mantisv2`, `clean_intermediates`).
3. Rebase this branch onto `main`, reconciling the `mantis-v2.nf` design against main's, and re-point QC at the templates from #298.

The QC-specific commits from `refactor-qc` (#295 reconcile, #297 standalone+templates) are already carried by #298 — they should NOT be re-applied here.


ieivanov:
I'll gather some notes here for us to resolve at the end:
* is `biahub_project` necessary? ideally we'd like to use the shared biautils conda environment; for testing, it's still useful to be able to use venvs. can we assume that the right biahub environment is already activated and call biahub functions directly, i.e. remove `biahub_project` and `biahub_cmd`? If not, can we leverage an environmental variable (e.g. BIAHUB=/path/to/env)
* we should aim to submit all jobs (maybe except for vs) as preempted - would preemption cause any problems with nextflow? from biahub perspective that's not an issue

aofei-liu:
@ieivanov apologies for the delay and thanks for your patience! nope, `biahub_project` was just my placeholder so the environment can be agnostic to e.g. whether it's conda or a uv or a pixi managed virtual env (since you just point it to the binary for whichever one) - if there's a particular one we should use for biahub, we can fix that

I think when i was prototyping the preemptions were getting reclaimed before the job finished (but this was with older flatfield/deskew code which was taking longer, so it might be that). the nextflow pipeline has retry logic currently baked in that will retry on larger compute on failure (to try and gracefully handle edgecases that oom slightly) - not sure how that interacts with preemption reclaims

ieivanov:
Closing as implemented.

PR #273, opened 2026-07-08
feat(nextflow): make per-step wall-time dynamic and volume-based

tayllatheodoro:
## Summary

Wall-time for the fan-out Nextflow steps was hardcoded (flat-field/deskew 60 min, reconstruct 360 min) and virtual-stain used a per-window budget that mis-fits both dataset geometries. Fixed times mis-provision in **both** directions across geometries: a long-thin timelapse (large T, small YX) and a short-wide acquisition (small T, large ZYX) need very different wall-times.

This extends `estimate_resources()` to also return a wall-time estimate. It now returns `(time_minutes, num_cpus, gb_ram_per_cpu)`, with a `time_multiplier` knob mirroring the existing `ram_multiplier`. Callers that only need CPUs and RAM use `_, num_cpus, gb_ram_per_cpu = estimate_resources(...)`.

Wall-time scales with the **number of ZYX volumes** processed (`T * C`), not total voxels. Per-voxel throughput is not a stable quantity — it depends on the CPU/GPU model, filesystem write speed, and chunking — whereas volume count is a property of the dataset alone. The per-volume spread between, say, an A549 volume and a neuromast volume is absorbed by `time_multiplier`, which is a fudge factor rather than a physical constant.

## The fixed budgets were killing jobs

Not just over-provisioning. Measured from `sacct` over ~3,300 step jobs since June 2026:

| Step | Jobs killed at the wall limit | Where |
|------|-------------------------------|-------|
| deskew | 15 / 761 (2.0%) | neuromast `2026_06_04` |
| flat-field | 8 / 1291 (0.6%) | A549 `2026_07_14`, `2026_05_27` |
| virtual-stain | 2 / 674 (0.3%) | neuromast `2026_06_04` |

These surface as `FAILED` rather than `TIMEOUT`, with `Elapsed == Timelimit` (e.g. `00:59:59 / 01:00:00`), which is why they were easy to miss.

## Calibration

Multipliers set to ~2x the worst per-volume rate observed across **6 A549 runs and 3 neuromast runs**:

| Step | time_multiplier | Worst observed | Margin across all 9 runs |
|------|-----------------|----------------|--------------------------|
| flat-field | 0.7 min/volume | 0.34 | 2.2 – 12.3x |
| deskew | 0.5 min/volume | 0.24 | 2.1 – 5.9x |
| reconstruct | 3.0 min/volume | 1.51 | 2.1 – 21.6x |
| virtual-stain | 5.0 min/volume | 2.53 | 2.0 – 5.7x |

Worked example (deskew): A549 gets 110 min against a 47.9 min observed max (2.3x); neuromast `48hpf` gets 660 min against 307.4 min (2.1x). Both near the 2x target, from a single multiplier.

## Why volumes and not voxels

Tested rather than assumed, using observed maxima to back out the implied multiplier under each key:

| Step | Dataset | Volumes | min/volume | min/Gvox |
|------|---------|---------|------------|----------|
| flat-field | A549 `07_01` | 201 | 0.250 | 0.549 |
| flat-field | A549 `07_14` | 201 | 0.340 | 0.747 |
| flat-field | neuromast `48hpf` | 1316 | 0.275 | 0.749 |
| | **spread** | | **1.36x** | **1.36x** |

A tie, so volumes wins on simplicity at no accuracy cost. Caveat: our two dataset families have similar per-volume sizes (4.55e8 vs 3.67e8 voxels), so this evidence cannot strongly discriminate the two keys — worth re-checking if we process a dataset with a very different ZYX.

## Virtual staining moved onto the same currency

The per-window (`T*Z`) budget fits poorly because a window's cost also depends on its YX footprint, which the window count ignores. Measured seconds/window:

- **A549: 1.29 – 1.58 s/window** (`07_01`, `06_16`, `05_27`)
- **neuromast: 0.31 s/window** (`48hpf`)

No single value covers both: 1.0 s/window under-provisions A549, and 5.0 over-provisions neuromast ~16x (9,377 min requested against 580 min used). Per-volume cost is far tighter (0.88 – 2.53 min/volume, ~2.9x spread), so virtual-stain now uses 5.0 min/volume and both families land at 2.0 – 5.7x.

## Known limitation: reconstruct

Reconstruct is the one step where no geometric key works. Per-volume cost spans **10x** across runs:

| Dataset | Volumes | Observed max | min/volume |
|---------|---------|--------------|------------|
| A549 `06_16` | 67 | 101.4 min | 1.513 |
| A549 `07_01` | 67 | 59.7 min | 0.891 |
| A549 `07_14` | 67 | 10.1 min | 0.151 |
| neuromast `06_04` | 266 | 146.4 min | 0.550 |
| neuromast `48hpf` | 658 | 91.8 min | 0.140 |

Keying on voxels does not help (also ~10x) — the reconstruction geometries are only 1.8x apart (A549 96x1664x1193 vs neuromast 171x768x794), so the variance comes from the reconstruction *settings* (regularization, TV iterations), not data size.

Consequence: sizing for the worst case means neuromast `48hpf` requests **21.6x** its observed max, against **3.9x** under today's hardcoded 360 — **for that dataset this PR is a regression.** Options:

1. Accept it (current state; tradeoff documented in-code).
2. Keep the hardcoded 360 for reconstruct only, taking the win on the other three steps.
3. Derive reconstruct's multiplier from the reconstruction settings.

Currently (1), for a uniform design. (2) is defensible.

## Notes

- `RESOURCES` contract unchanged.
- `main` merged in; `tests/test_cli/test_deskew_cli.py` (added in #278, after this branch forked) still unpacked the old 2-tuple and is fixed.
- CI green: Lint, build, Test on Python 3.12 and 3.13.

### DIFF
--- a/biahub/apply_inverse_transfer_function.py
+++ b/biahub/apply_inverse_transfer_function.py
@@ -25,7 +25,7 @@ from biahub.cli.parsing import (
 )
 from biahub.cli.utils import (
     echo_resources,
-    estimate_time_minutes,
+    estimate_resources,
     get_submitit_cluster,
     yaml_to_model,
 )
@@ -115,15 +115,15 @@ def apply_inverse_transfer_function(
     max_num_cpus = 16
     num_cpus, mem_per_cpu = wo_estimate_resources(list(input_shape), settings, max_num_cpus)
     mem_gb = num_cpus * mem_per_cpu
-    # Wall-time scales with total voxels reconstructed. Phase reconstruction only
-    # processes the input channels named in the config, so scale T*Z*Y*X by that
-    # count rather than the full C. Calibrated from a completed run: worst-case
-    # ~2.4 h for a single-channel 2.8e10-voxel volume -> ~3e6 voxels/s (this step
-    # is largely single-threaded per position). safety_factor covers the rest.
+    # CPUs/RAM come from waveorder above. Wall-time scales with total voxels
+    # reconstructed: phase reconstruction only processes the config's input
+    # channels, so pass a shape with C = that count and take only the time.
+    # time_multiplier ~= 11 min/Gvox calibrated from a completed run (~2.4 h for
+    # a single-channel ~2.8e10-voxel volume on ~1 CPU, with a ~2x margin).
     T, C, Z, Y, X = input_shape
     n_reconstructed = len(settings.input_channel_names)
-    time_minutes = estimate_time_minutes(
-        T * n_reconstructed * Z * Y * X, voxels_per_second=3.0e6
+    time_minutes, _, _ = estimate_resources(
+        (T, n_reconstructed, Z, Y, X), time_multiplier=11.0, max_num_cpus=max_num_cpus
     )
     echo_resources(num_cpus, mem_gb, time_minutes)
 
--- a/biahub/apply_inverse_transfer_function.py
+++ b/biahub/apply_inverse_transfer_function.py
@@ -25,6 +25,7 @@ from biahub.cli.parsing import (
 )
 from biahub.cli.utils import (
     echo_resources,
+    estimate_time_minutes,
     get_submitit_cluster,
     yaml_to_model,
 )
@@ -114,7 +115,16 @@ def apply_inverse_transfer_function(
     max_num_cpus = 16
     num_cpus, mem_per_cpu = wo_estimate_resources(list(input_shape), settings, max_num_cpus)
     mem_gb = num_cpus * mem_per_cpu
-    time_minutes = 360
+    # Wall-time scales with total voxels reconstructed. Phase reconstruction only
+    # processes the input channels named in the config, so scale T*Z*Y*X by that
+    # count rather than the full C. Calibrated from a completed run: worst-case
+    # ~2.4 h for a single-channel 2.8e10-voxel volume -> ~3e6 voxels/s (this step
+    # is largely single-threaded per position). safety_factor covers the rest.
+    T, C, Z, Y, X = input_shape
+    n_reconstructed = len(settings.input_channel_names)
+    time_minutes = estimate_time_minutes(
+        T * n_reconstructed * Z * Y * X, voxels_per_second=3.0e6
+    )
     echo_resources(num_cpus, mem_gb, time_minutes)
 
     if init_only:
--- a/biahub/apply_inverse_transfer_function.py
+++ b/biahub/apply_inverse_transfer_function.py
@@ -115,15 +115,22 @@ def apply_inverse_transfer_function(
     max_num_cpus = 16
     num_cpus, mem_per_cpu = wo_estimate_resources(list(input_shape), settings, max_num_cpus)
     mem_gb = num_cpus * mem_per_cpu
-    # CPUs/RAM come from waveorder above. Wall-time scales with total voxels
-    # reconstructed: phase reconstruction only processes the config's input
-    # channels, so pass a shape with C = that count and take only the time.
-    # time_multiplier ~= 11 min/Gvox calibrated from a completed run (~2.4 h for
-    # a single-channel ~2.8e10-voxel volume on ~1 CPU, with a ~2x margin).
+    # CPUs/RAM come from waveorder above. Wall-time scales with the number of
+    # volumes reconstructed: phase reconstruction only processes the config's
+    # input channels, so pass a shape with C = that count and take only the time.
+    #
+    # time_multiplier = 3.0 min/volume, ~2x the worst rate observed over
+    # completed runs (A549 2026_06_16, 101 min for 67 volumes = 1.51 min/vol).
+    # NOTE: reconstruct is the one step where per-volume cost is not
+    # geometry-driven -- it spans 0.14-1.51 min/volume across runs (10x), and
+    # keying on voxels instead is no better (also ~10x), because the cost is
+    # dominated by the reconstruction algorithm settings (regularization, TV
+    # iterations) rather than data size. Sizing for the worst case therefore
+    # over-requests on many-volume datasets; see PR #273 discussion.
     T, C, Z, Y, X = input_shape
     n_reconstructed = len(settings.input_channel_names)
     time_minutes, _, _ = estimate_resources(
-        (T, n_reconstructed, Z, Y, X), time_multiplier=11.0, max_num_cpus=max_num_cpus
+        (T, n_reconstructed, Z, Y, X), time_multiplier=3.0, max_num_cpus=max_num_cpus
     )
     echo_resources(num_cpus, mem_gb, time_minutes)
 
--- a/biahub/cli/utils.py
+++ b/biahub/cli/utils.py
@@ -456,5 +456,14 @@ def estimate_time_minutes(
     int
         Estimated wall-time in minutes (ceil), never below ``floor_minutes``.
     """
+    if num_voxels < 0:
+        raise ValueError("num_voxels must be >= 0.")
+    if voxels_per_second <= 0:
+        raise ValueError("voxels_per_second must be > 0.")
+    if floor_minutes < 0:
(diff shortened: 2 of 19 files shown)


ieivanov:
Thanks, this is helpful! Have you tested these timing estimates on reconstructions with cells too? I think I disagree with changing `seconds_per_window` to 1.0 in virtual staining

tayllatheodoro:
I'm currently using this branch to reconstruct the neuromast dataset, and it's going great! The virtual staining change is because the FOV in the neuromast is smaller than the cells, leading to an overestimation of the time by a large margin. I'll updated once all the steps finish

ieivanov:
Instead of a new `estimate_time_minutes` function, I'd suggest extending `estimate_resources`. The function can return `int(time_minutes), int(num_cpus), int(gb_ram_per_cpu)`. CLIs that don't take advantage of the time estimates can use `_, num_cpus, gb_ram_per_cpu = estimate_resources(...)`. I suggest using a `time_multiplier` to mirror our `ram_multiplier` instead of the current `safety_factor`.

ieivanov:
@tayllatheodoro let me know when you think this PR is ready and I can take a closer look

ieivanov:
@tayllatheodoro it's important that we merge this PR soon, I think it's a key improvement to the reconstruction pipeline. Please plan to test it with two datasets - neuromasts and cells, and share out the results here (namely time requested by time used; we should aim for requesting 2x more time than we estimate we'd need). 

I also think that specifying the time based on *total voxels* processed may be too detailed. This will depend on the exact cpu/gpu, writing speed, etc. I think it's better and simpler to estimate the time based on *total volumes* (i.e. XYZ) processed. The difference between processing a A549 volume vs neromast volume should be absorbed in the `time_multiplier` fudge factor. For example, for A459 recons we may request 2x more time than needed, whereas for neuromast recons we may request only 1.5x more time than needed - that's OK.

tayllatheodoro:
@ieivanov Tested on both dataset families as requested, and switched to volume-based scaling. Rebased… actually merged `main` in (branch was conflicting). Summary below; all numbers from `sacct` over ~3,300 step jobs since June — **6 A549 runs and 3 neuromast runs** — joined against input shapes read from the zarr stores.

## 1. The fixed wall-times were killing jobs

This is the part I hadn't appreciated before. It isn't only over-provisioning — the fixed 60 min budget times out on **both** families:

| Step | Jobs killed at the wall limit | Where |
|---|---|---|
| deskew | 15 / 761 (2.0%) | neuromast `2026_06_04` |
| flat-field | 8 / 1291 (0.6%) | A549 `2026_07_14`, `2026_05_27` |
| virtual-stain | 2 / 674 (0.3%) | neuromast `2026_06_04` |

These show up as `FAILED` (not `TIMEOUT`) with `Elapsed` == `Timelimit`, e.g. `00:59:59 / 01:00:00`, which is why they were easy to miss.

## 2. Voxels vs. volumes — you're right, switched to volumes

I tested this rather than assuming. Using per-step observed maxima and computing the implied multiplier under both keys:

| Step | dataset | volumes | min/volume | min/Gvox |
|---|---|---|---|---|
| flat-field | A549 `07_01` | 201 | 0.250 | 0.549 |
| flat-field | A549 `07_14` | 201 | 0.340 | 0.747 |
| flat-field | neuromast `48hpf` | 1316 | 0.275 | 0.749 |
| | **spread** | | **1.36x** | **1.36x** |

A dead tie, so volumes wins on simplicity at zero accuracy cost. One honest caveat: our two families have similar per-volume sizes (4.55e8 vs 3.67e8 voxels), so this data can't *strongly* discriminate the two keys — the tie is real but weakly evidenced. If we ever process a dataset with a very different ZYX, worth re-checking.

`estimate_resources` now returns `(time_minutes, num_cpus, gb_ram_per_cpu)` with a `time_multiplier` in **min/volume**, as you suggested.

## 3. Requested vs. used

Multipliers set to ~2x the worst observed per-volume rate:

| Step | min/volume | worst observed | margin across all 9 runs |
|---|---|---|---|
| flat-field | 0.7 | 0.34 | 2.2 – 12.3x |
| deskew | 0.5 | 0.24 | 2.1 – 5.9x |
| reconstruct | 3.0 | 1.51 | 2.1 – 21.6x ⚠️ |
| virtual-stain | 5.0 | 2.53 | 2.0 – 5.7x |

Worked example, deskew: A549 gets 110 min against a 47.9 min observed max (2.3x); neuromast `48hpf` gets 660 min against 307.4 min (2.1x). Both near the 2x target you asked for, from one multiplier.

## 4. Virtual staining — you were right about 1.0 s/window, but 5.0 isn't the fix either

Measured s/window from completed runs:

- **A549: 1.29 – 1.58 s/window** (`07_01`, `06_16`, `05_27`)
- **neuromast: 0.31 s/window** (`48hpf`)

So 1.0 s/window genuinely under-provisions A549 — your objection was correct, and my "smaller FOV" reply only explained the neuromast half. But reverting to 5.0 isn't right either: that's what produced the 9,377 min (156 h!) request against 580 min used on neuromast `48hpf`, a 16x over-request.

The reason no single value works is that a window's cost also scales with the **YX footprint**, which the T*Z window count ignores. Moving virtual-stain onto the same min/volume currency as the CPU steps tightens the spread from ~5x to ~2.9x and puts both families at 2.0–5.7x. **This is a change beyond what you asked for, so flagging it explicitly for your call.**

## 5. Reconstruct is a poor fit and this PR makes it worse for neuromast

Being upfront about this one. Per-volume cost spans **10x** across runs:

| dataset | volumes | observed max | min/volume |
|---|---|---|---|
| A549 `06_16` | 67 | 101.4 min | 1.513 |
| A549 `07_01` | 67 | 59.7 min | 0.891 |
| A549 `07_14` | 67 | 10.1 min | 0.151 |
| neuromast `06_04` | 266 | 146.4 min | 0.550 |
| neuromast `48hpf` | 658 | 91.8 min | 0.140 |

Keying on voxels instead does **not** help (also ~10x) — I checked, using the real reconstruction geometry (A549 96x1664x1193 vs neuromast 171x768x794, only 1.8x apart). The cost is driven by the reconstruction *settings* (regularization, TV iterations), not by data size, so no geometric key captures it.

Consequence: sizing for the worst case (3.0 min/volume) means neuromast `48hpf` requests **21.6x** its observed max, versus **3.9x** under today's hardcoded 360. **For that dataset this PR is a regression.** Options, your call:

1. Accept it (current state, tradeoff documented in-code).
2. Keep the hardcoded 360 for reconstruct only, and take the win on the other three steps.
3. Derive reconstruct's multiplier from the reconstruction settings.

I've gone with (1) so the design is uniform, but (2) is defensible and I'd switch on your word.

## Notes

- `RESOURCES` contract unchanged.
- Fixed a break the merge surfaced: `tests/test_cli/test_deskew_cli.py` (added in #278, after this branch forked) still unpacked the old 2-tuple.
- Full CLI suite: 56 passed, 1 skipped.
- The PR description above is now stale — it still describes `estimate_time_minutes` and the voxels/second table. Happy to rewrite it once we settle items 4 and 5.

tayllatheodoro:
Hey @ieivanov, could you review this branch? Its ready for merge

ieivanov:
This is excellent, thank you! If I understand correctly, `time_minutes` on derived from `estimate_resources` only for the 4 steps which have been calibrated so far - flat-field, deskew, reconstruct, and virtual-stain. Other recon steps still use hardcoded `time_minutes`, but the can be calibrated in the future. That sounds good to me.

Your decision on the timing of `reconstruct` sounds good - it's OK if we over-request to accommodate the worst case. Reconstruct uses a lot of Fourier transforms so it may strongly depend on what CPU we get, not just on the data size. In the future we may chose to exclude some old CPUs from the requests to speed up the reconstruction and decrease the variance in time per volume.

Merging now.

ieivanov (approved):

PR #289, opened 2026-07-24
fix(virtual-stain): suffix output channels with _prediction

### LINKED ISSUE #288: virtual-stain output channels drop the `_prediction` suffix (regression vs. viscy)
## Summary

`biahub virtual-stain` names its output channels **verbatim** from `cfg.data.init_args.target_channel`, whereas viscy's `HCSPredictionWriter` used to append `_prediction` to each target channel. With the same config, prediction stores now come out as `nuclei` / `membrane` instead of `nuclei_prediction` / `membrane_prediction`.

## Where

**biahub** — `biahub/virtual_stain.py` (`_init_output_plate`):

```python
create_empty_plate(
    store_path=output_dirpath,
    position_keys=[...],
    channel_names=target_channels,          # <-- used verbatim, no suffix
    shape=(T, len(target_channels), Z, Y, X),
    ...
)
```
where `target_channels = cfg.data.init_args.target_channel`.

**viscy (old behavior)** — `viscy-utils/src/viscy_utils/callbacks/prediction_writer.py`:

```python
prediction_channel = [ch + "_prediction" for ch in target_channel]
```

So a config with `target_channel: ["nuclei", "membrane"]` produced `nuclei_prediction` / `membrane_prediction` under viscy, but produces plain `nuclei` / `membrane` under biahub.

## Why it matters

The `_prediction` suffix is an implicit contract that biahub's own downstream code relies on:

- `biahub/settings.py` defaults `input_channel` / `target_channel` to `"nuclei_prediction"`
- `biahub/track.py` `mem_nuc_contour(nuclei_prediction, membrane_prediction)`
- viscy's airtable `filter_raw_channels` treats `nuclei_prediction` / `membrane_prediction` as virtual stains

So existing configs (which correctly list the GT channel name `nuclei`/`membrane` as `target_channel`) now yield stores that don't match what the rest of the pipeline expects, and the channels must be renamed by hand after the fact.

## Suggested fix

Restore the viscy naming contract in `virtual_stain.py` by deriving output names from the target channels, e.g.:

```python
prediction_channels = [f"{ch}_prediction" for ch in target_channels]
```
and use `prediction_channels` for `create_empty_plate(channel_names=...)`.

Alternatively, make the suffix behavior explicit/opt-out via config if verbatim naming is sometimes desired — but the default should match viscy to avoid silently breaking downstream steps.

## Observed on

- Repo: `czbiohub-sf/biahub`, branch `refactor-track` @ `a58bcd3`
- Affected dataset (worked around by manual rename): `2026_07_14_A549_MAP4_ZIKV.zarr`

tayllatheodoro:
## Summary

`biahub virtual-stain` named its output channels **verbatim** from `cfg.data.init_args.target_channel`, dropping the `_prediction` suffix that viscy's `HCSPredictionWriter` used to append. With the same config, prediction stores came out as `nuclei` / `membrane` instead of `nuclei_prediction` / `membrane_prediction`, and had to be renamed by hand.

This restores the viscy naming contract by concatenating `_prediction` onto each target channel in `_init_output_plate`:

```python
channel_names=[f"{ch}_prediction" for ch in target_channels],
```

## Why

The `_prediction` suffix is an implicit contract the rest of the pipeline relies on:

- `biahub/settings.py` defaults `input_channel` / `target_channel` to `"nuclei_prediction"`
- `biahub/track.py` `mem_nuc_contour(nuclei_prediction, membrane_prediction)`
- viscy's airtable `filter_raw_channels` treats `nuclei_prediction` / `membrane_prediction` as virtual stains

The `cytoland` provenance metadata still records the original `target_channel` (`nuclei`/`membrane`) — correct, since that's the model's training target; only the output channel *labels* get the suffix. The write path indexes positionally, so no other code changes are needed.

Closes #288

### DIFF
--- a/biahub/virtual_stain.py
+++ b/biahub/virtual_stain.py
@@ -255,10 +255,13 @@ def _init_output_plate(
         scale = input_dataset.scale
     T, C, Z, Y, X = input_shape
 
+    # Match viscy's HCSPredictionWriter naming contract: output channels are the
+    # target channels suffixed with "_prediction" (e.g. "nuclei" -> "nuclei_prediction").
+    # Downstream biahub steps (settings, track, airtable) key off this suffix. See #288.
     create_empty_plate(
         store_path=output_dirpath,
         position_keys=[Path(p).parts[-3:] for p in input_position_dirpaths],
-        channel_names=target_channels,
+        channel_names=[f"{ch}_prediction" for ch in target_channels],
         shape=(T, len(target_channels), Z, Y, X),
         scale=scale,
         version=resolve_ome_zarr_version(input_position_dirpaths[0], output_ome_zarr_version),


ieivanov:
Fixed in #305

PR #290, opened 2026-07-24
feat(tile_stitch): honour monarch.recon_dtype as the stored output dtype

srivarra:
## What

`create_stitch_output` hardcoded `np.float32`, so a `recon_dtype: float16` run still wrote a **float32** store — double the output bytes on disk. `ReconDtype`'s own docstring says *"Dtype the recon is **stored** + RDMA-transmitted in"*, so this implements documented-but-missing behaviour rather than adding a knob.

- `_output_dtype(config)` derives the store dtype from `monarch.recon_dtype` (`StrEnum` of numpy dtype names, so it converts directly).
- `create_stitch_output` uses it in both the plate and FOV branches.
- `_chunk_shape` now takes `itemsize` instead of assuming 4 bytes, so a narrower dtype fits **more z per chunk** (float16: z 242 → 484, both landing at 63.9 MB) rather than emitting chunks at half the 64 MB target.

Default is unchanged (`ReconDtype.FLOAT32`) — existing configs are unaffected.

## Why it matters here

For the 2026_04_07 DaXi timelapse recon, output is `(2368, 1436, 4445)` per (timepoint, channel). At float32 that's **56.3 GiB** each; 64 timepoints × 2 channels ≈ **7.2 TiB** of intermediate recon output. float16 halves it, which materially changes what fits on `/hpc/projects/waveorder`.

## Tests

- `test_output_store_honours_recon_dtype` — parametrised over unset / `float32` / `float16`, asserts the store dtype and chunk agreement.
- `test_chunk_shape_scales_leading_extent_with_itemsize` — asserts f16 doubles the z extent and stays under the cap.
- Existing `_chunk_shape` assertions updated to pass `itemsize` explicitly (kept required rather than defaulted — a wrong default silently halves chunk sizes).

Chunk math and lint verified. The full suite needs the `feature/device-transfer-api` waveorder that this base branch pins, which isn't installed in my environment.

## ⚠️ Unrelated blocker on this base branch

`feature/tile-stitch-dataflow` repins waveorder to `feature/device-transfer-api`. That branch forked from `feat/anisotropic-yx-pixel-size` at `0d85183` — **one commit before** `f9cbd7e "pass in pinhole for fluorescence reconstructions (#570)"` — and it is the *only* commit it lacks.

Consequence: on this base branch `confocal_pinhole_diameter` has **0 pass-throughs**, so it is silently ignored and fluorescence reconstructs **widefield**. Cherry-picking `f9cbd7e` onto `feature/device-transfer-api` fixes it.

### DIFF
--- a/biahub/tile_stitch/orchestration.py
+++ b/biahub/tile_stitch/orchestration.py
@@ -64,12 +64,26 @@ def _is_plate_position(path: Path) -> bool:
         return False
 
 
-def _chunk_shape(tile_spatial: tuple[int, ...]) -> tuple[int, ...]:
-    """Keep float32 TCZYX chunks near 64 MB for parallel zarrs writes."""
+def _output_dtype(config: TileStitchReconSettings):
+    """Return the stored output dtype, taken from ``monarch.recon_dtype``.
+
+    ``ReconDtype`` is a ``StrEnum`` of numpy dtype names, so it converts directly.
+    """
+    import numpy as np
+
+    return np.dtype(str(config.monarch.recon_dtype))
+
+
+def _chunk_shape(tile_spatial: tuple[int, ...], itemsize: int) -> tuple[int, ...]:
+    """Keep TCZYX chunks near 64 MB for parallel zarrs writes.
+
+    ``itemsize`` follows the stored dtype, so a float16 run gets chunks with twice the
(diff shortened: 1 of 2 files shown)


srivarra:
need to go punish claude code. unnecessary pr.

PR #296, opened 2026-07-30
fix(registration): complete AntsRegistrationSettings so the ants config path runs

tayllatheodoro:
Bug fix. `biahub estimate-registration` with `estimation_method: ants` has never been runnable from a config.

## The bug

`estimate_tczyx` reads four fields off `AntsRegistrationSettings`:

```python
# biahub/registration/ants.py:507-510
crop=ants_registration_settings.crop,
ref_mask_radius=ants_registration_settings.ref_mask_radius,
clip=ants_registration_settings.clip,
sobel_filter=ants_registration_settings.sobel_filter,
```

The model defined only `sobel_filter`, so the config-driven path raised `AttributeError` before reaching ANTs. Only direct `estimate_czyx` calls worked, which is how this went unnoticed.

## The fix

Add `crop`, `ref_mask_radius` and `clip`, with defaults matching `preprocess_czyx`'s signature so config-driven and direct calls behave identically when unset.

`ref_mask_radius` is validated at config-parse time. `preprocess_czyx` already rejects out-of-range values, but only after the data is loaded — failing on the config is much cheaper.

The `clip` docstring records a trap worth knowing: its limits are hardcoded for a **phase** reference (`np.clip(ref, 0, 0.5)`), so it flattens anything else — a virtual-staining prediction ranges well above 0.5.

## Why no test caught it

There were example configs for `beads` and `manual` but none for `ants`, and `test_all_example_settings_tested` only checks that every *existing* file is tested. The untested path was invisible.

This adds the missing example config plus two regression tests that keep the model and its consumer in sync:

- `test_ants_settings_cover_estimate_tczyx_reads` greps `estimate_tczyx` for `ants_registration_settings.<field>` and asserts the model covers every one, so adding a read without a field fails the suite.
- `test_ants_settings_defaults_match_preprocess_czyx` asserts the defaults agree with the function that consumes them.

Both were verified to **fail** against the old model rather than pass vacuously.

## Scope

Three files, additive. No existing QC or registration behaviour is touched — zero diff against `QCBeadsRegistrationSettings`, `EvalTransformSettings`, `evaluate_transforms`, `validate_transforms` and `overlap_score`. Every `settings.py` change is inside the `AntsRegistrationSettings` class.

34 settings tests pass, ruff clean.

### DIFF
--- a/biahub/settings.py
+++ b/biahub/settings.py
@@ -258,7 +258,46 @@ class AffineTransformSettings(MyBaseModel):
 
 
 class AntsRegistrationSettings(MyBaseModel):
+    """Settings for the ANTs registration backend.
+
+    Field names and defaults mirror the keyword arguments of
+    ``biahub.registration.ants.preprocess_czyx``, which is what consumes them.
+
+    Attributes
+    ----------
+    sobel_filter : bool
+        Apply a Sobel filter (3D gradient magnitude) to both volumes before
+        registering, so ANTs matches structural edges rather than raw
+        intensity. Needed for cross-modality pairs such as fluorescence
+        against a virtual-staining prediction.
+    crop : bool
+        Crop both volumes to their overlapping region with the LIR algorithm
+        before registering.
+    ref_mask_radius : float | None
+        Radius of a circular mask applied to the reference channel, as a
+        fraction of image width in ``(0, 1]``. ``None`` applies no mask.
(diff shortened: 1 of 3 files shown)


ieivanov (approved):

ieivanov:
Thanks!

Issue #206, opened 2026-02-27
pip install error with pyqt6


tayllatheodoro:
```(biahub_test) [taylla.theodoro@gpu-sm02-06 biahub]$ pip install -e .
Obtaining file:///hpc/mydata/taylla.theodoro/repo/biahub
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Requirement already satisfied: iohub<0.4,>=0.3.0a5 in /hpc/mydata/taylla.theodoro/anaconda/2022.05/x86_64/envs/biahub_test/lib/python3.11/site-packages (from iohub[tensorstore]<0.4,>=0.3.0a5->biahub==0.1.0) (0.3.0a6)
Collecting stitch @ git+https://github.com/ahillsley/stitching@jen (from biahub==0.1.0)
  Cloning https://github.com/ahillsley/stitching (to revision jen) to /tmp/pip-install-cfx7lzod/stitch_dfbaf28001284d34b65d7463438215dc
  Running command git clone --filter=blob:none --quiet https://github.com/ahillsley/stitching /tmp/pip-install-cfx7lzod/stitch_dfbaf28001284d34b65d7463438215dc
  Running command git checkout -b jen --track origin/jen
  Switched to a new branch 'jen'
  branch 'jen' set up to track 'origin/jen'.
  Resolved https://github.com/ahillsley/stitching to commit aefca1ce87c288d98cc976a180e80e9c218e5982
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: matplotlib in /hpc/mydata/taylla.theodoro/anaconda/2022.05/x86_64/envs/biahub_test/lib/python3.11/site-packages (from biahub==0.1.0) (3.10.8)
Requirement already satisfied: napari in /hpc/mydata/taylla.theodoro/anaconda/2022.05/x86_64/envs/biahub_test/lib/python3.11/site-packages (from biahub==0.1.0) (0.6.6)
Collecting PyQt6 (from biahub==0.1.0)
  Using cached pyqt6-6.10.2.tar.gz (1.1 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
[...]

PR #200, opened 2026-02-18
feat: Integrate mantis-analysis-template as reusable pipeline templates

tayllatheodoro:
Merges the mantis-analysis-template repository into biahub as a collection of reusable analysis pipeline templates. Includes three template variants for different microscopy workflows.

### 1. Mantis Template (templates/mantis/)
Complete pipeline for standard Mantis microscope workflows:
- Zarr conversion and data management
- Label-free phase reconstruction with waveorder
- Virtual staining with VisCy
- Light-sheet deskewing and deconvolution
- Multi-channel registration (manual/beads/ANTs)
- XYZ stabilization
- Cell tracking with ultrack
- Visualization and QC tools

### 2. Dragonfly Template (templates/dragonfly/)
Pipeline variant optimized for Dragonfly spinning disk confocal systems
- Similar structure to Mantis template
- Dragonfly-specific configurations and parameters

### 3. Zebrafish Physics Pipeline (templates/zebraphysics-mantis-pipeline/) Specialized pipeline for zebrafish developmental biology experiments:
- Extended drift analysis tools
- Physics-specific data processing
- Customized tracking parameters
- Additional SLURM batch configurations

### Template CLI Commands
```bash
# List available templates
biahub template list

# Show template details
biahub template info mantis

# Initialize new project from template
biahub template init -t mantis -o ./my-project --dataset exp_001
```

### Template Structure
Each template includes:
- **0-convert/**: Data conversion and setup
- **1-preprocess/**: Label-free and light-sheet preprocessing
- **2-assemble/**: Data assembly and concatenation
- **3-visualization/**: Visualization and QC scripts
- **run_pipeline.py**: Pipeline orchestration
- **check_logs.py**: SLURM log monitoring
- **README.md**: Complete documentation

## Files Added

- `templates/` - New templates directory with all three templates (~250 files)
- `biahub/cli/template_manager.py` - Template management CLI
- `templates/README.md` - Templates overview and usage guide
- `MANTIS_TEMPLATE_INTEGRATION.md` - Detailed integration documentation

## Files Modified

- `biahub/cli/main.py` - Added template CLI commands

### Copy Template for New Project
```bash
cp -r biahub/templates/mantis ~/my-experiment
cd ~/my-experiment
export DATASET=my_dataset_name
python run_pipeline.py
```

### Using Template CLI
```bash
# Initialize new project
biahub template init -t mantis -o ~/my-experiment --dataset exp_2026

# List templates
biahub template list

# Get template info
biahub template info mantis
```

[...]

PR #198, opened 2026-02-13
Preserve zarr version in register command

mattersoflight:
Add version parameter to create_empty_hcs_zarr() and detect input zarr version in register_cli() to ensure output version matches input (v2→v2, v3→v3). This maintains compatibility with both OME-NGFF v0.4 (zarr v2) and v0.5 (zarr v3) datasets.

Changes follow the pattern from waveOrder PR#519:
- Add optional version parameter to create_empty_hcs_zarr (default "0.4")
- Pass version to open_ome_zarr when creating output plate
- Detect input version from source plate in register_cli
- Backward compatible: existing calls use default "0.4"

Pair-coded-with: Soorya Pradeep <@Soorya19Pradeep>

PR #298, opened 2026-08-03
feat(qc): QC integrated into mantis-v2, with imaging-qc as a direct dependency

### LINKED ISSUE #306: feat(nextflow): make pipeline steps optional (skip tracking for neuromast/zebrafish)
## Problem

`nextflow/mantis-v2.nf` runs a fixed chain: flat-field → deskew → reconstruct → virtual-stain → assemble → track. Every step is mandatory. The workflow body errors without each config:

```groovy
if (!params.track_config)       error "Provide --track_config"
if (!params.concatenate_config) error "Provide --concatenate_config"
```

and then unconditionally wires the step:

```groovy
track_trigger = assemble_done.done
track_wf(all_positions, track_input, track_input_images, track_output, params.track_config, track_trigger)
```

**Tracking is an A549 step.** For neuromast/zebrafish datasets the deliverable is `5-assemble`; tracking is not part of what those runs are for, and the tracking parameters (cellpose `diameter`, `min_area`/`max_area`, linking `max_distance`) are tuned for A549 cells, not neuromasts.

There is no way to express that today. A zebrafish run has to:

1. author or copy a `track.yml` it does not want,
2. pass it to `--track_config` purely so the workflow validates and starts,
3. spend GPU wall time on a `4-track` fan-out after assemble,
4. produce a `4-track/<ds>.zarr` that is discarded.

Since [#282](https://github.com/czbiohub-sf/biahub/pull/282) / [#279](https://github.com/czbiohub-sf/biahub/pull/279) reordered track to run *after* assemble and read the assembled plate, this is strictly serial: the unwanted step delays nothing else, but it does extend the run past the point where the deliverable already exists.

## What I'd like

A way to declare which steps a run performs. Rough shape, not a strong opinion on syntax:

```bash
--steps flat_field,deskew,reconstruct,virtual_stain,assemble        # neuromast
--steps flat_field,deskew,reconstruct,virtual_stain,assemble,track  # A549 (default)
```

with the config requirement following from the selection — `--track_config` required only if `track` is in the list. Nextflow's native `-entry` is the other obvious option, but it needs one named entry workflow per useful prefix and does not compose as well as a list.

Two things worth preserving:

- **`directory_layout()` stays the single source of truth.** A skipped step should not renumber the ones after it — `5-assemble` should stay `5-assemble` whether or not track runs, so paths remain comparable across datasets.
- **Validate the selection up front.** Related: [#304](https://github.com/czbiohub-sf/biahub/issues/304) asks for all step `init`s to run at the start so config errors surface immediately rather than hours in. Step selection should be validated in the same early pass — including rejecting a selection whose first step's input does not exist.

## Why it matters beyond zebrafish

The same gap shows up whenever the chain needs to stop early or restart partway:

- re-running only `assemble` after fixing a concatenate config, without re-validating the upstream configs
- stopping after `reconstruct` to eyeball phase before committing GPU hours to virtual staining
- [#292](https://github.com/czbiohub-sf/biahub/issues/292)'s verify-then-clean, which wants assemble and track decoupled

## Workaround today

Pass a placeholder `track.yml` and discard `4-track`. This is what the `reconstruct-with-nextflow` skill does for zebrafish ([PR #305](https://github.com/czbiohub-sf/biahub/pull/305), `templates/configs/zebrafish/track.yml`), with the placeholder marked as such and the plan telling the user the output is a by-product. Once this lands, that placeholder and the caveat around it can be deleted.

aofei-liu:
QC runs as part of a reconstruction. `mantis-v2.nf` gains `--qc_config` (assembled store) and `--qc_track_config` (tracking store); both become tabs of **one** report at `${output}/qc/report`. `qc-standalone.nf` remains the entrypoint for QC'ing already-reconstructed datasets, and produces the same unified report from a manifest.

Supersedes #322, whose branch was merged into this one.

imaging-qc-pipeline#199 has **merged**, so `imaging_qc_pipeline` is pinned to a commit on imaging-qc `main` (`2be2ee6`) rather than a branch tip. That pin brings in Phases 20 through 22 — see below.

## The environment contract

`qc_cmd()` prefixed every QC task with `uv run --from 'imaging-qc-pipeline @ git+…'`, so up to `maxForks` tasks re-resolved an environment and re-fetched a git dependency concurrently against one shared site-packages. QC was the last holdout after `biahub_cmd()` went away; it now follows the same contract as the rest of the pipeline — resolve once, call `imaging-qc` bare:

```bash
uv sync --extra qc
source .venv/bin/activate
```

The `qc` extra is deliberately **not** in `all`: imaging-qc-pipeline is a private repo with no PyPI release, and CI's `uv sync --group dev` resolves `all` with no credentials for it. `check_environment()` now takes the tools the caller actually runs, so a run without QC is not held to an `imaging-qc` it never calls.

## Where QC sits in the pipeline

QC is a labelled reconstruction step: it joins `reconstruction_steps()` when switched on, so the Slack run-start announcement lists it and per-step counters read N/7 rather than N/6. One Slack message covers the whole QC step — the unified report is a single task gated on every store's finalize, so its completion is exactly "QC of all stores is done". The run-end message already fires after QC without wiring, since `onComplete` runs at session teardown once every task in the DAG has finished.

`directory_layout()` is untouched: QC writes tables inside the stores it reads and its report to `${output}/qc/report`, so it owns no step directory.

### Scheduling

Each store waits only on the step that wrote it — the assembled plate on `assemble_done` (the same signal `track_wf` waits on, so image QC runs **concurrently with tracking**), the tracking store on `track_done`. Adding tracking QC serialises nothing that was parallel.

A QC verdict cannot fail the run: `imaging-qc gate` exits 0 whether positions pass or fail (measured), recording the verdict in each store's `tables/qc/` tables and a `QC_SUMMARY` line. Only a broken config or a genuine compute error exits non-zero.

## The report spec is written at launch, in Nextflow

`qc_report_spec()` in `modules/qc.nf` is a plain function — not a process, not a `biahub` CLI call. Its content is fully determined before any compute runs, from params and the pipeline's own directory layout, so a malformed spec fails at launch rather than after every compute task has finished (which is how the two bugs in the old generator were found). `qc_dir` is *declared* as each store's own `tables/qc/` group rather than probed on disk, since the pipeline passes no `--output-dir`.

`biahub/qc.py`, its tests and the `generate-report-spec` command are gone.

## Following imaging-qc through four upstream changes

The pin moved three times during this work, and each move was checked by re-running the subset comparison before anything else:

- **Phase 20 (#198)** removed the wave mechanism. plan.json is v5, one flat `items[]`; `consolidate --wave-id`, `scope:` on a metric group and `finalize_wave` are gone. The driver follows: one fan-out, one merge.
- **Phase 21 (#200)** trimmed the registry 15 → 11. None of the removed metrics is one we name, so no verdict of ours moves.
- **Phases 22/22b/22c** rework the report. Measured gain: the report's `plots/` went 34.06 MB → 4.96 MB, because fragments now share one `plotly.min.js` instead of embedding 4.85 MB each; `_tmp_plots/` is cleaned up; the store table's compute status reaches the Dataset Overview.
- **Two fixes we asked for landed**, and both local workarounds are now deleted: [#201](https://github.com/czbiohub-sf/imaging-qc-pipeline/issues/201) (`report --config <file>` now resolves the file's Hydra defaults — so the spec names the config file, `3322b87`) and [#205](https://github.com/czbiohub-sf/imaging-qc-pipeline/issues/205) (upstream pins `wasmtime<48`, so our `constraint-dependencies` entry is gone, `adad216`).

## Fan-out, and where the parallelism is

- **`qc_chunk_size` 10 → 0**: one compute task per position. imaging-qc threads a position's `(T, C)` units internally, sized from the task's CPU affinity, so a position is already parallel inside its node. Chunking T only multiplies jobs — 100 positions × T=100 is **100 tasks at 0, 1000 at 10**.
- **`compute_step` gains `cpus`** (`qc_cpus`, default 8). It had none, so that thread pool saw one CPU and walked a position's timepoints serially.

## Config migration, and the per-step directories

Pinning imaging-qc exposed four keys the current version refuses or ignores, all found by running the configs rather than reading them: `bleach_rate` (no longer a metric), `blank.gate.aggregation` (refused — imaging-qc's changelog names this file by path), `granularity: z` (retired for `slice`), `report.mode` (ignored). No threshold or metric-set changes.

Configs are grouped **one directory per pipeline step**, not by stage number:

```
nextflow/configs/qc/assemble/{base,pixel_metrics}.yaml
nextflow/configs/qc/track/cell_count.yaml
```

That grouping is what makes each tab resolve its own report block. imaging-qc scans the directory a tab is given and takes the first YAML with a `report:` section, so one shared directory made both tabs resolve to `base.yaml` — the cell-count tab rendered pixel metric labels and no `instance_count` plot, at exit 0. A per-step directory cannot, whatever the stage numbers are. The stage number now appears only in each config's `stage:` key, so renumbering a stage is that one edit, with no files to rename.

Related upstream finding, filed as [imaging-qc-pipeline#201](https://github.com/czbiohub-sf/imaging-qc-pipeline/issues/201): `report --config <file>` does not compose Hydra `defaults:`, so a stage config inheriting its `report:` block loses every metric plot. Hence the spec points each tab at the config *directory*.

## Tested end to end on a real reconstruction

`2026_08_11_A549_SEC61B_DENV` through `mantis-v2.nf` with QC on, 54 positions, T=67, Z=86, 6 channels:

```
/hpc/projects/intracellular_dashboard/organelle_dynamics/2026_08_11_A549_SEC61B_DENV_debug/v3/
```

Both stores were QC'd and both are tabs of one report at `v3/qc/report/index.html` (9.4 MB, 7 plots):

| store | stage | compute | positions | pass | fail |
|---|---|---|---|---|---|
| `4-assemble` | 5 | ok | 54 | 38 | 16 |
| `5-track` | 6 | ok | 54 | 44 | 10 |

The directory numbering is the new positional scheme in production: `4-assemble` and `5-track`, with QC owning no directory of its own. The report spec was written at launch and names each tab's config file.

Per store, QC produced the consolidated metric table (`stage5_position.parquet`, 5,882,868 rows at `(position, T, C, Z, metric_name)` grain — 85 MB), the per-position gate verdicts, the per-`(position, T)` drop list (3,618 rows, 48 failing), and the one-row store summary. Compute was 54 tasks per store at `qc_chunk_size = 0`, and the whole QC stage took ~25 min of the run.

**Two earlier attempts failed, and both failures are fixed in this PR** — that is the substance of the test, not an aside:

1. **A 7h29m run aborted after every reconstruction step had finished**, in the QC wiring: `combine` concatenated the trigger, so tracking's collected list of 30 paths turned a 3-tuple into a 32-tuple and no closure matched. Fixed in `cba08ee` by mapping the trigger instead. Reproduced in eight lines of Nextflow first; `-preview` cannot catch it, because it builds the DAG without executing operators.
2. **`finalize_stage` then exited 1** on a config error — imaging-qc refused the per-channel SNR `hard_floor` because its keys named channels the store does not have. The floors were keyed `raw GFP EX488 EM525-45`, but the `raw ` prefix only appears after the manual `rename_channels.py` step, so the bound would have flagged nothing. Fixed in `6464b06`. The guard did exactly its job: a silently-inert gate was refused rather than passing everything.

The Slack message for failure (1) was also worse than the bug — `"reconstruction aborted: Possible solutions: any(), any(), …"`, Groovy's suggestion list, for a run where 288 tasks succeeded and no task failed. `cba08ee` fixes the title to name the failure kind and the real error line, and strips stack frames from the detail.

Earlier, `qc-standalone.nf` was validated against the 2026-08-12 DynaCell QC run on an index-aligned subset: **metric values reproduce exactly**, 360/360 stage-5 rows and 134/134 stage-6 rows identical, max delta 0. Full write-up, configs and comparison script:

```
/hpc/projects/intracellular_dashboard/organelle_dynamics/qc_runs_debug_2026-08-18_current_pipeline/
```

## Channel coverage

Metrics run on every channel in a store unless a config restricts them, so the assembled store is QC'd across all 6 channels (the 2026-08-12 run restricted to 3; that restriction is deliberately not carried over).

One consequence worth knowing: imaging-qc's store-level summary aggregates across channels, and prediction channels are near-binary, so their SNR is enormous — `nuclei_prediction` mean 4500 versus 19 for GFP. The aggregate therefore reads ~880 rather than the ~17 of the raw channels. Per-channel `hard_floor` gates are unaffected (they key by name), and every per-slice value is in the tables; it is the single summary number that is dominated by the derived channels.

## Before merge

- [ ] **Calibrate the gate thresholds.** The SNR floor (5.0) and tenengrad floor (0.001) are generic starting points, and on this dataset 16 of 54 positions fail stage 5. Someone who knows the experiment has to decide which of those are real before a failing verdict means anything.
- [ ] **Decide what the store-level SNR summary should be.** Metrics run on every channel unless a config restricts them, which is intended — but imaging-qc's summary aggregates across channels, and the near-binary prediction channels dominate it (`nuclei_prediction` mean SNR 4500 vs 19 for GFP), so the single number reads ~880 where the raw channels alone read ~17. Per-channel gates are unaffected. Either ask upstream for a per-channel summary or accept the number as meaningless.
- [x] **Give `uv-lock.yml` a credential** for the private imaging-qc repo, or accept that the weekly refresh fails. PR CI and docs CI need nothing: `uv sync --group dev` installs 297 packages and none is imaging-qc, because the `qc` extra is deliberately outside `all`. Needs a biahub admin — `maintain` cannot manage secrets, and a deploy key needs admin on the imaging-qc side.
- [x] **Merge the branch up from `main`** — one commit behind (`4e3095d`, a notify CA-bundle fix).
- [x] Optional: now that #201 is fixed, the per-step config *directory* grouping is no longer load-bearing for report resolution, since the spec names the file. Flattening `configs/qc/{assemble,track}/` back to one directory would work; keeping it also documents which store each config is for.
- [ ] Optional: `qc_cpus = 8` is still a guess. This run gives the first real numbers to tune it against.

Related upstream question, not a blocker: [#206](https://github.com/czbiohub-sf/imaging-qc-pipeline/issues/206) asks whether a T-spanning metric (track length) is in scope, or whether a driver should compute it. The same gap covers a max-over-Z gate — the Z collapse is a hard-coded mean with no reducer knob.


ieivanov:
@aofei-liu I think I'm reasonably happy with this PR now, let me know what you think.

One possible blocker - https://github.com/czbiohub-sf/imaging-qc-pipeline is a private repo which makes it a bit hard to depend on. Users would need authentication to build the environment, I added info in the `reconstruct-with-nextflow` skill to help with that. Another option would be for us to maintain a wheel on bruno until imaging-qc-pipeline becomes public. I'm looking to make the installation as smooth as possible, let me know what you think

ieivanov:
@aofei-liu could you please take one more look thru this PR? If the code looks good to you, we can merge it and make smaller changes from there

aofei-liu:
@ieivanov yup lgtm! I think i'll need to bump the version (which might change the schema slightly) to address some of the concerns you had yesterday during our meeting, but for now nothing seems to be breaking

ieivanov (approved):

PR #299, opened 2026-08-04
Feat/resume interrupted positions

ieivanov:
## Problem

The per-position steps run on the `preempted` partition, so SLURM reclaims tasks mid-write. Two things went wrong when it did:

1. **The retry got stuck.** A killed write left a torn shard, which every later attempt read back and rejected — the position never completed and the run aborted.
2. **The retry threw away good work.** Even once unstuck, a retry recomputed the position from scratch, discarding every (t, c) unit the killed attempt had already finished.

Assemble was the worst case: one job copying the whole plate — 26.5k units and ~16 TiB of I/O against a 6 hour limit — so a kill near the end discarded all of it.

Separately, every `--init` step took minutes and wrote GBs, on a step whose only job is to create an empty plate.

### Resume instead of restart

`flat-field`, `deskew`, and `concatenate` gain `--resume`, threaded through to `process_single_position`; the Nextflow per-position tasks opt in. A retry now recomputes only the units the killed attempt had not finished.

The completion record is keyed by a fingerprint of the **resolved settings** (`settings_fingerprint`), so re-running a step with a changed config against an existing output store recomputes rather than silently skipping units whose data would now differ. `--resume` is off by default — only the Nextflow tasks turn it on, where "recompute what this position is missing" is always what's wanted.

For concatenate, the three source stores write disjoint channels of each output position, so their units never collide.

Verified by killing an assemble partway (7 completion records dropped, plus one shard torn under a record that still claimed done): the resume recomputed exactly those 8 units, skipped the other 16, and produced output bit-identical to an uninterrupted run.

### Stop copying acquisition metadata into every derived store

`create_empty_plate(metadata_sources=...)` copied **every** non-OME zattrs key from the input plate. The mantis acquisition writer leaves an `ome_writers.frame_metadata` list with one record per raw Z-frame — 214,668 entries, 18.5 MB compact, 39 MB once written back — and that rode from the raw store into every derived plate. iohub then wrote it one key at a time, re-reading the destination for each membership test, making the copy quadratic in the metadata size.

Measured on `2026_07_14_A549_MAP4_ZIKV_rerun` (72 positions):

| init step | wall | written |
|---|---|---|
| flat-field | 3m 09s | 4 GB |
| deskew | 6m 14s | 4 GB |
| reconstruct | 7m 28s | 6 GB |
| concatenate | 14m 38s | 10 GB (16 GB read) |
| virtual-stain — *the one step with no `metadata_sources`* | **37s** | **632 KB** |

Each step now passes an explicit allowlist (`PROVENANCE_METADATA_KEYS = ("biahub-*", "waveorder", "cytoland")`) rather than inheriting whatever an upstream store happens to carry. The blob is stale by then anyway: its frame indices describe the raw 1068-plane Z stack, not the 86-plane deskewed output. `normalization` is dropped alongside it — it describes the reconstruction step's inputs, not its output.

Result: `deskew --init` over 66 positions goes **6m 14s → 49s** (most of the remainder is interpreter startup), the per-position `zarr.json` drops from **39 MB → 2.7 KB**, and the array and OME metadata are byte-identical to the store built by the current code.

## iohub dependency

This needs unreleased iohub, pinned to `czbiohub-sf/iohub#dev/biahub-pipeline` — a short-lived branch combining two features, each reviewable on its own:

- **`feat/interrupt-safe-writes`** — replaces a shard torn by a preempted job instead of reading it back (the recurring "checksum is invalid" abort), and adds the per-unit resume this branch drives via `--resume`.
- **`feat/filter-copied-position-metadata`** ([iohub#454](https://github.com/czbiohub-sf/iohub/pull/454)) — adds `create_empty_plate(metadata_keys=...)`.

Repoint at each feature branch as it merges, then at the PyPI release. **This PR should not merge before those do.**

## Notes for reviewers

- **Reconstruct is not covered by resume.** It drives its own per-timepoint loop in waveorder, so it gets the torn-shard repair from iohub but not per-unit resume.
- **No graceful-shutdown path.** An earlier commit on this branch (`b10b1b0`) described SIGTERM letting the in-flight write finish before stopping. That was measured upstream to never fire under Nextflow — the generated `.command.run` traps TERM, kills the payload, and exits without waiting, and slurmstepd follows with SIGKILL — so it has since been removed from iohub. The protection that actually does the work is repair plus resume, both kept.
- **Resume records now live beside the store**, in a `.iohub-progress/` sibling of the output zarr rather than inside the array directory. Copying or deleting a store therefore no longer carries or clears its progress state, and a `.iohub-progress/` directory will appear next to each output plate.

## Testing

`pytest tests/` — 126 passed, 1 skipped.

Beyond that, this branch produced a full 341-task pipeline run end to end (`2026_07_14_A549_MAP4_ZIKV_rerun`), which absorbed two real preemptions via retry.

### DIFF
--- a/biahub/apply_inverse_transfer_function.py
+++ b/biahub/apply_inverse_transfer_function.py
@@ -24,6 +24,7 @@ from biahub.cli.parsing import (
     sbatch_to_submitit,
 )
 from biahub.cli.utils import (
+    PROVENANCE_METADATA_KEYS,
     echo_resources,
     estimate_resources,
     get_submitit_cluster,
@@ -68,6 +69,7 @@ def _init_output_plate(
         position_keys=[Path(p).parts[-3:] for p in input_position_dirpaths],
         **output_metadata,
         metadata_sources=input_plate,
+        metadata_keys=PROVENANCE_METADATA_KEYS,
     )
 
     return input_shape, channel_names
--- a/biahub/cli/parsing.py
+++ b/biahub/cli/parsing.py
@@ -305,6 +305,25 @@ def monitor() -> Callable:
     return decorator
 
 
+def resume() -> Callable:
+    def decorator(f: Callable) -> Callable:
+        return click.option(
+            "--resume/--no-resume",
+            "resume",
+            default=False,
+            show_default=True,
+            help=(
+                "Skip the (time, channel) units this position already finished in an "
+                "earlier attempt instead of recomputing the whole position. For retrying "
+                "a run that was interrupted, e.g. by Slurm preemption. A finished unit is "
+                "skipped without re-deriving it, so pass --no-resume (or use a fresh "
+                "output store) when the settings changed."
+            ),
+        )(f)
+
+    return decorator
+
+
 def num_processes() -> Callable:
     def decorator(f: Callable) -> Callable:
         return click.option(
--- a/biahub/cli/utils.py
+++ b/biahub/cli/utils.py
@@ -1,3 +1,4 @@
+import hashlib
 import json
 import logging
 import os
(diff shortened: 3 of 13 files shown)


ieivanov:
TODO:
- [x] transfer metadata during virtual stain
- [ ] develop fast concatenate
- [ ] don't hardcode time required for concatenate
- [ ] enable resume for waveorder reconstruct? have waveorder use iohub process_single_position rather than its own implementation? shard repair still works.
- [x] VS does not do repair or resume - no needed since it runs on gpu partition?
- [ ] don't retry with increased resources? - helps get around frozen nodes, requires decent estimation of resources
- [ ] ~~flatfield should not copy channels it does not process~~
- [ ] restructure nextflow DAG to run positions in parallel; init steps should all run at the very start to validate configs

PR #300, opened 2026-08-04
perf(concatenate): place chunk files directly when the data is unchanged

ieivanov:
Stacked on #285's branch (`feat/resume-interrupted-positions`) — review that first; the diff here is the four files below.

## What

`biahub concatenate` assembled its output by reading each input volume, cropping it, and writing it back — a full decompress/recompress of every byte. When the output geometry matches the input and no ROI crop is requested, that round trip produces stored files that are byte-for-byte what the input already holds, so the file can be placed at its new key instead.

Measured on real `5-assemble` shards (4.9 GiB, production overhang geometry, channel remapped), read back and compared element-wise against the read-write path:

| path | time |
|---|---|
| direct copy | **1.7 s** |
| read-write (`process_single_position` + `copy_n_paste`) | 33.3 s |

19.5×, and that is with the bytes crossing Lustre to NFS. With `--link` on one filesystem it is a metadata operation.

## How it decides

A stored chunk is interpretable only through its array's metadata — data type, chunk grid, codecs, fill value, key encoding. `copy_incompatibilities` therefore compares the **whole metadata document** rather than a hand-picked list of fields, so a codec option added by a future Zarr release disqualifies the copy by default instead of silently producing an output that cannot be decoded. On top of that:

- ZYX shape must match and the Z/Y/X slices must be the full extent of the input;
- the `(t, c)` remap must be file-aligned — one channel and one timepoint per file, which is what a default `shards_ratio` gives (T ratio 1, and `process_single_position` already rejects sharding along C). A multi-timepoint shard is allowed only when the copy is the identity on the time axis.

Anything else falls back to the read-write path **per position**, with the reason printed, so a mixed plate still gets the speedup where it applies:

```
Direct chunk copy: 4/4 position sources (18 files)
  read-write fallback (2): array metadata differs: data_type (input 'uint16', output 'float32'), codecs
```

## Two side effects worth knowing

- **Placement is atomic** — written under a temporary name in the final directory, then `os.replace`d. The torn-shard failure the base branch added repair and resume for cannot occur on this path at all.
- **The memory request drops** when every source is copyable: nothing buffers a shard, so `mem_gb` is capped at 8 instead of the shard-sized `estimate_resources` value. For the mantis-v2 assemble step that is 96 GB → 8 GB today, and 816 GB → 8 GB if it moves to the 10-timepoint shards discussed in `HANDOFF_torn_shard_resume.md` §7 — which removes the objection that it would only schedule on the 44 large `preempted` nodes.

## One deliberate behaviour change

A byte copy cannot apply `copy_n_paste`'s `nan_to_num`, so **NaNs are carried through rather than zeroed**. Documented on the CLI flag and in the module docstring. A block the input never wrote still has no file and the destination is left with none, so it reads back as the fill value exactly as before; a stale destination file with no source is *removed*, matching what a skipped write leaves behind.

`resume` on this path means "the destination file is already the same size as its source" rather than iohub's per-unit markers. Exact for a run interrupted here, since placement is atomic; a file torn by an interrupted *read-write* run is only caught if its size differs.

## Flags

- `--direct-copy` / `--no-direct-copy` (default on) — opt out to take the read-write path everywhere.
- `--link` (opt-in) — hard-link instead of copying bytes. Instant and free of disk, but the output then shares data with its inputs, so writing into either would corrupt the other. Only safe for write-once inputs. Downgrades to a byte copy across filesystems rather than failing with `EXDEV`.

## Not in this PR

`nextflow/modules/assembly.nf` needs no change to get the speedup, since `--direct-copy` is on by default. Adding `--link` there would make assemble near-instant and zero-disk, and is arguably safe because the sources are write-once and deleted after assemble is verified — but it means the assembled plate shares inodes with `1-deskew`/`2-reconstruct`/`3-virtual-stain` until those are removed. Left as a separate decision.

## Tests

`tests/test_direct_copy.py` (23) covers layout reading for v0.4 and v0.5, each disqualifying condition, and the placement mechanics (link mode sharing inodes, missing source clearing a stale destination, resume, no leftover scratch files). Three integration tests in `tests/test_concatenate.py` assert the direct and read-write outputs are element-wise equal, in both copy and link mode, and that a Z crop falls back. Full suite: 152 passed, 1 skipped. `pre-commit` clean.

### DIFF
--- a/biahub/cli/direct_copy.py
+++ b/biahub/cli/direct_copy.py
@@ -1,0 +1,729 @@
+"""Place Zarr chunk/shard files directly, for copies that do not change the data.
+
+``biahub concatenate`` normally assembles its output by reading each input
+volume, cropping it, and writing it back — a full decompress/recompress of every
+byte. When the output geometry matches the input exactly and no crop is
+requested, that round trip produces stored files that are byte-for-byte what the
+input already holds, so the file can simply be placed at its new key instead.
+For the mantis-v2 assemble step that is ~16 TiB of codec work replaced by a file
+copy, or by nothing at all in ``link`` mode.
+
+Two properties make the decision mechanical rather than a judgement call:
+
+*The array metadata document is the contract.* A stored chunk or shard is
+interpretable only through its array's metadata — data type, chunk grid, codecs,
+fill value, key encoding. If two arrays agree on all of it a file is
+interchangeable between them, and if they disagree anywhere it is not.
+:func:`copy_incompatibilities` therefore compares the whole document instead of
+a hand-picked list of fields, so a codec option introduced by a future Zarr
+release disqualifies the copy by default rather than silently producing an
+output that cannot be decoded.
+
+*The file is the unit.* One file holds one chunk-grid cell — a shard when the
+array is sharded, a chunk when it is not. Remapping ``(t, c)`` is then a rename
+of that cell, but only if a cell holds exactly one channel and one timepoint,
+which is what the pipeline's stores do (``shards_ratio`` defaults to a T ratio
+of 1, and ``process_single_position`` already rejects sharding along C).
+Anything coarser is rejected.
+
+Placement is atomic: each file is written under a temporary name in its final
+directory and then ``os.replace``d into place. An interrupted run therefore
+leaves either the old file or the new one and never a torn one, which is the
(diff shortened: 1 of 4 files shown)


ieivanov:
Closing in favor of #337

Issue #291, opened 2026-07-27
Normalize channel names in the assembly step (BF, Phase3D, *_prediction, raw prefix)


tayllatheodoro:
## Problem

After the mantis pipeline assembles a plate (`assemble_wf` → `concatenate`), the channel names in the output store are whatever the upstream sources happened to emit:

- `BF - Oblique` from the label-free/deskew store
- `Phase3D...` (sometimes with a suffix) from reconstruct
- `nuclei` / `membrane` from virtual stain
- raw fluorescence channels named by the microscope filter set, e.g. `mCherry EX561 EM600-37`, `GFP EX488 EM525-45`

This makes every downstream step (segmentation, tracking, feature extraction, dashboards) depend on acquisition-specific strings, and the names differ between experiments. I've been fixing this up by hand with a post-hoc script in my `5-assemble/` directory, and I think it belongs in the pipeline instead.

## Naming convention I've been applying

| Incoming | Renamed to |
| --- | --- |
| `BF - Oblique` | `BF` |
| `Phase3D*` | `Phase3D` |
| `nuclei` | `nuclei_prediction` |
| `membrane` | `membrane_prediction` |
| anything else (e.g. `mCherry EX561 EM600-37`) | `raw <original name>` |

The `raw ` prefix marks channels that are unprocessed detector output, so `raw`-prefixed and derived channels are distinguishable at a glance. The rule is idempotent — re-running never yields `raw raw mCherry ...`.

## Script I'm currently running

`DATASET=<plate-name-without-.zarr>` in the cwd, then:

```python
import os
from pathlib import Path

from iohub import open_ome_zarr

dataset = os.environ.get("DATASET")
if dataset is None:
    print("DATASET environmental variable is not set")
    exit()
dataset_path = f"{dataset}.zarr"
# Check if the dataset exists
if not Path(dataset_path).exists():
    raise FileNotFoundError(f"Dataset {dataset_path} does not exist.")

# Channels to leave as-is (everything else gets a 'raw ' prefix)
keep_channels = [
    "BF",
    "Phase3D",
    "nuclei_prediction",
    "membrane_prediction",
]

# Open the dataset and rename channels
with open_ome_zarr(dataset_path, mode="a") as ds:
    for pos_name, pos in ds.positions():
        print(f"Processing position: {pos_name}")
        for channel_name in ds.channel_names:
            if channel_name == "BF - Oblique":
                print(f"Renaming channel '{channel_name}' to 'BF'")
[...]

PR #305, opened 2026-08-06
feat(skills): add reconstruct-with-nextflow Claude skill

### LINKED ISSUE #308: nextflow: per-task `uv run` mutates the shared venv (30 concurrent writers); pass --no-sync
## Summary

Every per-task `uv run --project ...` invocation reinstalls a package into the shared `.venv`. With `maxForks = 30` that is up to 30 concurrent processes writing the same `site-packages` during a run.

## Measured

Three consecutive invocations, nothing else touching the environment:

```
$ uv run --project /hpc/mydata/taylla.theodoro/repo/biahub biahub nf --help
Uninstalled 1 package in 54ms
Installed 1 package in 232ms
# again
Uninstalled 1 package in 47ms
Installed 1 package in 220ms
# again
Uninstalled 1 package in 42ms
Installed 1 package in 206ms
```

It never converges — `uv run` syncs the project environment on every invocation, and something (a git-sourced dependency) is re-laid-down each time.

`--no-sync` eliminates it completely:

```
$ uv run --no-sync --project ... biahub nf --help   # x2
no mutation
no mutation
```

## Where

`nextflow/modules/common.nf`:

```groovy
def biahub_cmd() {
    return params.biahub_project ?
        "uv run --project ${params.biahub_project} biahub" : "biahub"
}
```

`nextflow/modules/virtual_stain.nf`:

```groovy
def stain_cmd(tool) {
    return params.biahub_project ?
        "uv run --project ${params.biahub_project} --extra stain ${tool}" : tool
}
```

Note these ask uv for **different environment specs** against one shared venv: every step except virtual-stain requests the default set, virtual-stain requests `+ stain`. So consecutive tasks are asking uv to reconcile the same directory to two different targets.

## Why it matters

- **Concurrency.** 30 tasks × uninstall-then-install into one `site-packages` is an unsynchronised read-modify-write on shared state. A task importing a module while another is mid-reinstall is a plausible source of the kind of transient `ImportError`/`ModuleNotFoundError` that is otherwise very hard to attribute.
- **It contradicts "don't touch the checkout during a live run."** We tell people not to edit `$BIAHUB_PROJECT` because it invalidates `-resume`; meanwhile the tasks themselves mutate its venv on every invocation.
- **Lustre.** Every mutation is a package uninstall + reinstall on a shared filesystem that cannot hardlink from the uv cache (`Failed to hardlink files; falling back to full copy`), so each one is a full copy.

## Suggested fix

Add `--no-sync` to both command builders, so tasks only ever *read* the environment:

```groovy
"uv run --no-sync --project ${params.biahub_project} biahub"
"uv run --no-sync --project ${params.biahub_project} --extra stain ${tool}"
```

or set `env.UV_NO_SYNC = '1'` in `nextflow.config` so it applies to every task uniformly.

This makes provisioning the environment an explicit, deliberate step before launch rather than an implicit side effect of the first task, which is what we want anyway.

## Related, and a caution

Provisioning that environment is **not** as simple as `uv sync`. This checkout's `.venv` has **27 packages that are not in `uv.lock`** — `cupy-cuda12x`, `tracksdata`, `stitch`, `dexp`, `dask-cuda`, `dask-image`, `distributed`, `ilpy`, `pyscipopt`, `polars`, `pims` and their closure — and uv correctly considers them extraneous:

```
$ uv sync --dry-run             # Would uninstall 33 packages
$ uv sync --all-extras --dry-run # Would uninstall 28 packages
```

I removed them by running `uv sync` and had to restore them by hand. On Lustre it also fails partway (`failed to remove directory ... __pycache__: Directory not empty (os error 39)`), and a half-completed uninstall left `cupy`'s `dist-info` without a `RECORD`, after which the stale directory shadowed the real module and `iohub` / `cytoland` / `ultrack` all failed with `AttributeError: module 'cupy' has no attribute 'ndarray'`.

Worth deciding separately whether those 27 should be declared in `pyproject.toml` (so the environment is reproducible from the lockfile) or whether the venv is intentionally hand-managed and `uv sync` should simply never be run against it. Right now it is neither documented nor safe.

tayllatheodoro:
## What

Adds a Claude Code skill at `.claude/skills/reconstruct-with-nextflow/` that drives a mantis-v2 dataset through `nextflow/mantis-v2.nf` on Bruno — flat-field → deskew → reconstruct → virtual-stain → assemble → track.

Invoked as `/reconstruct-with-nextflow dataset <DATASET_NAME>`.

## Why

The knowledge needed to run this pipeline currently lives in per-project `run_mantis_v2.sh` copies, one-off `build_plate*.py` scripts, and handoff markdown scattered across `/hpc/projects`. Each new dataset means rediscovering the same caveats, and some of them (VS `z_window_size` mismatch, the migrated `TrackingSettings` schema) fail silently or fail hours in. This puts that in one reviewable, version-controlled place.

## What it does

1. Confirms the shell is on Bruno; runs the Nextflow head process on a login node (it only submits SLURM jobs), with a compute-node reservation as the documented alternative.
2. Locates the raw acquisition under `/hpc/instruments/cm.mantis/<DATASET>` and **flags any deviation from the `<DATASET>_1.ome.zarr` convention rather than guessing** — see below.
3. Detects whether the store is a real HCS plate or a flat `{R}Pos{C}` acquisition needing a `0-convert` plate.
4. Proposes the output project directory from the dataset family, and **presents a plan for approval before writing anything**.
5. Scaffolds `configs/` and `run_mantis_v2.sh` in the output directory, preferring a copy from the most recent successful run of the same family, diffed against shipped templates.
6. Launches in a detached tmux session with the pipeline in the foreground, so the user attaches and watches; the skill reads the tee'd log rather than stealing the pane.
7. Monitors `squeue`, `.nextflow.log`, `trace.txt`, and the per-step SLURM logs; notifies on completion, terminating error, and excessive retries.
8. Renames channels on the assembled plate, verifies the store, and reports the Nextflow summary.

## Caveats it encodes

Each of these has cost a real run:

- **No HCS plate on neuromast/zebrafish/dynatrack acquisitions.** They are flat `bioformats2raw.layout: 3` stores; `list_positions` fails. Needs a `0-convert` symlink plate first — delegated to the existing `build-hcs-plate` agent.
- **Channel renaming after assemble** (`BF - Oblique` → `BF`, `raw ` prefix on acquired fluorescence), from the canonical script in `2026_05_27_A549_SEC61B_TOMM20_G3BP1_ZIKV`. The shipped copy refuses to double-prefix.
- **VS `z_window_size` must equal `in_stack_depth`** — a mismatch does not raise, the model just stalls. Two checkpoints in use (neuromast 21 / A549 15); both paths verified to exist.
- **`TrackingSettings` rejects `mode` / `z_range` / `focus_config`.** Several configs under `/hpc/projects/tlg2_mantis/` still carry the old schema; the mapping to `output_mode` / `z_slicing` is documented.
- **`concatenate.yml` placeholders stay placeholders** — the assemble subworkflow injects the real paths.
- **SLURM `.out`/`.err` look inverted** and `.command.log` is empty, per the deliberate crossing in `slurm_logs()`.
- Don't edit the biahub checkout during a live run (invalidates `-resume`); `uv lock && uv sync` once before launching.

## Error handling

Triage table in `references/recovery.md`, but the split that matters:

- **Exit 130–145** — routine preemption on the `preempted` partition. Nextflow already retries. Do nothing.
- **Lustre EIO / `RuntimeError: the checksum is invalid` / torn shard** — a killed task left a truncated shard; because every array overhangs in ZYX, every write is a read-modify-write, so this fails *identically on every retry and every `-resume`*. Routed to the existing `job-io-error-repair` agent. The rule it enforces: delete `<pos>/0/c/` only, keep both `zarr.json` files — the scaffold comes from a separate cached `init-*` task that `-resume` won't re-run.
- **Exit 1/2 with a traceback** — real bug or bad config; Nextflow terminates deliberately.

Sourced from `2026_07_14_A549_MAP4_ZIKV_rerun/HANDOFF_torn_shard_resume.md`, including the finding that the graceful-drain SIGTERM handler is inert under Nextflow.

## Notifications

`templates/notify.sh` posts to `$BIAHUB_SLACK_WEBHOOK` and falls back to printing in the terminal when unset, exiting 0 either way so a monitoring loop is never broken by a missing webhook. The URL is a credential — kept out of the repo.

## Verification

- All 12 config templates parse against the current settings classes: `DeskewSettings`, `FlatFieldCorrectionSettings`, `ConcatenateSettings`, `TrackingSettings`, waveorder's `ReconstructionSettings`, and both `virtual_stain.yml` files through `build_predict_parser()`.
- Both VS checkpoint paths exist on disk.
- `bash -n` on both shell templates; `ruff check` and `ruff format --diff` clean on `rename_channels.py`.
- `notify.sh` fallback path exercised.

No changes to any pipeline or library code — this is additive under `.claude/`.

## Worth a second opinion

- ~~The dataset-family → output-root rule is inferred from where past runs landed.~~ **Confirmed by Taylla:** `/hpc/projects/tlg2_mantis` for neuromast/zebrafish, `/hpc/projects/intracellular_dashboard/organelle_dynamics` for cells — which is what the skill already encodes. Instrument-QC datasets (argolight/beads/PSF/alignment) still have no default and the skill asks; say if they should.
- ~~`templates/configs/zebrafish/track.yml` has a flagged conflict between `nuclei` and `nuclei_prediction`.~~ **Resolved via #288:** `biahub virtual-stain` names outputs verbatim from `target_channel`, dropping the `_prediction` suffix viscy used to append, while the rest of biahub keys off the suffix. Both VS templates now carry the suffix in `target_channel` (the same workaround the A549 configs already use), so track resolves at track time and the rename becomes a no-op safety net.
- ~~The zebrafish track template defaults to `z_slicing: {method: focus, window_size: 40}`.~~ **Moot:** zebrafish is not tracked (below).

## Updated after review feedback

Three changes in `abdf286` / `25b67ee`:

**Renaming now follows #291, not the old per-project script.** The convention is `BF - Oblique` → `BF`, `Phase3D*` → `Phase3D`, `nuclei` → `nuclei_prediction`, `membrane` → `membrane_prediction`, `raw ` prefix on anything unmatched. Implemented with an explicit passthrough branch, so it is genuinely **idempotent** — the first version guarded against double-prefixing by refusing to run, which is weaker. Mapping and idempotency are unit-checked (12 cases, fixed-point verified).

**An ordering constraint the issues exposed, now documented.** The rename runs *after* the pipeline, but `4-track` runs *inside* it — so track reads the assembled plate under **pre-rename** channel names. Both VS templates carry `_prediction` in `target_channel` so this resolves correctly today.

**Zebrafish is not tracked.** The deliverable for neuromast/zebrafish is `5-assemble`. `mantis-v2.nf` hard-requires `--track_config` and unconditionally wires `track_wf`, so a zebrafish run still passes a config and still produces a `4-track` store — the template is now a marked placeholder, and the skill states in the plan that the output is a discarded by-product. The A549 tracking parameters in it are explicitly called out as *not* tuned for neuromasts.

Filed **#306** to make pipeline steps optional; once it lands the placeholder and its caveat come out. `templates/rename_channels.py` similarly goes away when #291 / PR #260 / PR #250 land a `rename-channels` step — the skill says to check for that before running the manual script.

All 12 config templates re-validated after the changes.

---

## Tested end-to-end on real data

Ran the skill against `2026_08_05_dynatrack_2dpf` — a zebrafish/dynatrack acquisition with three stores; used the T=2 / 8-position one as a smoke test before committing to the 5.4 TB timelapse.

**The reconstruction chain passed.** `flat_field`, `deskew`, `compute_transfer_function`, `apply_inv_tf`, `virtual_stain_preprocess`, `virtual_stain` and `concatenate` all COMPLETED for all 8 positions, first attempt, no retries. `5-assemble` came out correct:

```
positions: 8    shape: (2, 6, 171, 768, 794) float32    28 GB
channels : ['BF - Oblique', 'GFP EX488 EM525-45', 'mCherry EX561 EM600-37',
            'Phase3D', 'nuclei_prediction', 'membrane_prediction']
all channels finite; VS predictions non-trivial
```

The `_prediction` suffix arrived as intended, confirming the #288 workaround in the VS templates. The rename then applied the #291 convention and was verified **idempotent on the real store** (second run: "all channels already canonical — nothing to do"). `-resume` was verified too: changing only `track.yml` re-ran track alone and cached everything upstream.

The skill's naming check earned its keep immediately — the directory is `..._2dpf` but the stores are `..._2df_*` (missing `p`), with three of them, so it stopped and asked instead of guessing.

**Five defects found, all now fixed or filed.** Three were mine:

1. **`uv sync` guidance was destructive** — and I broke the environment by following it. The `.venv` carries **27 packages not in `uv.lock`** (`cupy-cuda12x`, `tracksdata`, `stitch`, `dexp`, `dask-cuda`, `ilpy`, `pyscipopt`, `polars`, `pims` + closure); every sync variant deletes them, it fails partway on Lustre with `ENOTEMPTY`, and the half-finished `cupy` uninstall left a `dist-info` without a `RECORD` whose stale directory then shadowed the module — cascading into `iohub`/`cytoland`/`ultrack` all failing to import. Restored from the pins uv printed and verified (27 intact, cupy 14.0.1 + CUDA, all six CLIs loading, `uv lock --check` clean). Replaced with verification that mutates nothing: `caveats.md` §11.
2. **zebrafish `track.yml` was broken as shipped** — `z_slicing: {method: focus, focus_channel: Phase3D}` while `input_images` loaded only `nuclei_prediction`. `apply_focus_slicing` resolves the focus channel against the *loaded* channels, not the store, so all 8 positions failed 5/5 retries **after** the whole reconstruction had succeeded. Now `method: all`, with the constraint documented.
3. **`monitoring.md` position count was wrong** — `find -mindepth 3` is one level short of `<store>.zarr/<row>/<col>/<fov>` (returned 2, not 8), and a bare `*.zarr` overcounts `2-reconstruct` because of `transfer_function.zarr`. Also now records that it counts init-created scaffolds, not completed work — a step whose every task failed still reads 8/8.

Two are biahub bugs, filed:

- **#308** — every per-task `uv run --project` reinstalls a package (measured 3/3, never converging), so up to 30 concurrent writers on one `site-packages`; `--no-sync` fixes it. Also asks whether those 27 packages should be declared or the venv treated as hand-managed.
- **#309** — `cellpose_config.gpu: true` hands **cupy** labels to ultrack's `labels_to_contours` → skimage → `scipy.ndimage`, which are CPU-only: `TypeError: Implicit conversion to a NumPy array is not allowed`. Plus a separate operational problem: the failing task drops into `(Pdb)` *inside the SLURM job*, holds its GPU allocation to the wall limit, and since a time-limit kill is retryable, Nextflow retries a deterministic failure 6×. Added to `recovery.md` since `errorStrategy` cannot detect it.

Neither blocks the zebrafish deliverable — tracking is not part of it (#306), which this run demonstrates: the pipeline reported FAILED purely because of a step whose output is discarded.

Corrections to figures cited earlier: store 3 is **5.4 TB on disk** (862 GB/position), not the ~19.9 TiB logical size; and `2026_08_05_dynatrack_2dpf` is newer than the datasets in the original survey.

---

## Environment handling reworked — supersedes defect 1 above

The `uv sync` guidance in "Five defects found" has been **superseded**. The churn had a concrete root cause — `uv.lock` recorded a `viscy-transforms` version the wheel from that same pinned commit does not declare — fixed with `uv lock --refresh-package viscy-transforms`. The 27-package hazard was specific to one checkout's out-of-band installs, not to `uv sync` itself.

The pipeline no longer wraps tasks in `uv run` at all: `--biahub_project`, `biahub_cmd()` and `stain_cmd()` are removed, and the environment is resolved once before launch and inherited through `sbatch --export=ALL`. Verified end-to-end on real data (16 tasks, 0 failures, 0 retries) with zero env mutation in any task log.

Full write-up, including a request to @tayllatheodoro about the out-of-band packages on her checkout: https://github.com/czbiohub-sf/biahub/pull/305#issuecomment-5258295436

Also fixes `--max_positions`, which aborted every run that used it (CLI params arrive as Strings; `take` has no String overload).

Closes #308


ieivanov:
I'm taking notes here as I'm reading this PR

>Scaffolds configs/ and run_mantis_v2.sh in the output directory, preferring a copy from the most recent successful run of the same family, diffed against shipped templates.

Once the configs are part of the shrimpy repo, it would be good if Claude opens a PR to update the configs with a motivation for why the change was needed. New runs should target configs on main  by default, configs from a previous run can be a fallback option.

>No HCS plate on neuromast/zebrafish/dynatrack acquisitions. They are flat bioformats2raw.layout: 3 stores; list_positions fails. Needs a 0-convert symlink plate first — delegated to the existing build-hcs-plate agent.

iohub now reads `bioformats2raw.layout` zarr layout, we should update `list_positions` such that it can extract positions from that zarr structure too. We shouldn't need a 0-convert shim for zebrafish datasets

> Lustre EIO / RuntimeError: the checksum is invalid / torn shard

With the features in https://github.com/czbiohub-sf/iohub/pull/455, writing should be robust to Lustre EOI. First try a rerun with --resume, then propose plan for repair instructing the user which chunks to delete. Don't delete data through this skill (dangerous)

>Zebrafish is not tracked. The deliverable for neuromast/zebrafish is 5-assemble. mantis-v2.nf hard-requires --track_config and unconditionally wires track_wf, so a zebrafish run still passes a config and still produces a 4-track store

We should make some configs optional (e.g. concatenate and track). If these are not passed, then these and downstream dependent Nextflow steps won't run, as suggested in #306

TODO:
- [x] move config files to repo
- [ ] update CLAUDE.md with instructions to update this skill with every PR merged to main
- [x] remove `BIAHUB_PROJECT` from nextflow arguments
- [x] do we need to cd into the output folder? where is the work folder saved? should we explicitly pass `-work-dir`?
- [x] Change invocation form `/reconstruct-with-nextflow dataset <DATASET_NAME>` to `/reconstruct-with-nextflow <DATASET_NAME>`?

ieivanov:
@tayllatheodoro can you please include instructions here on setting up $BIAHUB_SLACK_WEBHOOK? I'm curious to try it

.claude/skills/reconstruct-with-nextflow/templates/run_mantis_v2.sh
```diff
@@ -0,0 +1,70 @@
...
+#   0-flatfield/ 1-deskew/ 2-reconstruct/ 3-virtual-stain/ 4-track/ 5-assemble/
+# The work dir defaults to <output>/nextflow/work; override with `-work-dir`.
+#
+# Any extra arguments are forwarded to nextflow, e.g.
+#   ./run_mantis_v2.sh -profile local
+#   ./run_mantis_v2.sh --max_positions 2      # quick smoke test
+
+module load nextflow
+module load uv
+module load cuda/12.8.0_570.86.10
```
ieivanov: Why is `module load cuda/12.8.0_570.86.10` required here? Would `module load cuda` work? I've run the pipeline successfully even without that call - is it strictly necessary? Is it that the tracking run on CPU without it?

.claude/skills/reconstruct-with-nextflow/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+
+The store is normally `<DATASET_NAME>_1.ome.zarr` (Micro-Manager appends an
+acquisition index). **The convention is `YYYY_MM_DD_<description>` and the store
+name usually matches the directory name — when it does not, say so explicitly
+and ask which store to use before continuing.** This is common: e.g. directory
+`2026_08_04_smart_fov_selection_test` contains
+`2026_08_04_test_fov_selection_*.ome.zarr`, and multiple numbered/`_epi`/
+`_prescan`/`_fov_debug` variants sit side by side. Never silently pick one.
+
+Confirm the store is real, not an aborted acquisition:
```
ieivanov: Don't do `du -sh /hpc/instruments/cm.mantis/<DATASET>/<STORE>.ome.zarr` - it can take a while and doesn't guarantee that the acquisition was not aborted. Instead, when there are acquisitions with the same name and several acq indices, e.g. `DATASET_1.ome.zarr`, `DATASET_2.ome.zarr`, `DATASET_3.ome.zarr`, pick the one with the largest index.

.claude/skills/reconstruct-with-nextflow/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+|---|---|
+| zebrafish / neuromast / dynatrack | `/hpc/projects/tlg2_mantis/<DATASET>` |
+| A549 / cell-line / organelle | `/hpc/projects/intracellular_dashboard/organelle_dynamics/<DATASET>` |
+
+If the target directory already exists with step outputs in it, that is a
+**resume or a rerun** — say which you think it is and let the user decide before
+touching anything. The established convention for a fresh reprocess of the same
+data is a `<DATASET>_rerun` sibling.
+
+## 5. Find a reference run to copy configs from
```
ieivanov: See https://github.com/czbiohub-sf/biahub/pull/305#issuecomment-5247344846, we should prefer the configs on main

.claude/skills/reconstruct-with-nextflow/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+tmux new-session -d -s "$SESSION" -c "<OUTPUT>"
+tmux send-keys -t "$SESSION" "bash ./run_mantis_v2.sh 2>&1 | tee -a nextflow/run_$(date +%Y%m%dT%H%M%S).log" Enter
+```
+
+Tell the user: `tmux attach -t nf_<DATASET>` to watch, `Ctrl-b d` to detach.
+
+**Do not edit the biahub checkout while a run is live** — it changes Nextflow
+task hashes and invalidates `-resume`.
+
+**Never run `uv sync` on the biahub checkout.** 27 packages in its `.venv` are not
```
ieivanov: We should fix that - it's a good idea to run uv sync before every reconstruction run. If packages we need are not in .venv, then we should add them to pyproject.toml

.claude/skills/reconstruct-with-nextflow/references/caveats.md
```diff
@@ -0,0 +1,313 @@
...
+
+## 6. Assemble is a single job on one reserved node
+
+`concatenate --cluster debug` iterates every position in-process, so `5-assemble`
+is one large SLURM job rather than a fan-out — 16 TiB of I/O and tens of
+thousands of write units for a typical plate. A late kill used to discard all of
+it; `biahub concatenate --resume` (which the Nextflow task passes) makes it
+recoverable. Give it room: it is the single longest step.
+
+## 7. Shard geometry and `shards_ratio`
```
ieivanov, line 172: We should develop an iohub feature such that shards can be auto-sized to ~ 2GB by lumping several timepoints into one shard. channels are currently not allowed to be sharded

.claude/skills/reconstruct-with-nextflow/references/caveats.md
```diff
@@ -0,0 +1,313 @@
...
+the `clusterOptions` override wins over Nextflow's own `-o`. Read the
+`slurm_output` files, not the work dir logs.
+
+## 10. Do not edit the biahub checkout during a live run
+
+Editing files in `$BIAHUB_PROJECT` changes Nextflow task hashes and invalidates
+`-resume`, so a restart recomputes everything. Get the branch right *before*
+launching, then leave it alone.
+
+## 11. NEVER run `uv sync` on the biahub checkout — verify instead
```
ieivanov: Find the root cause and fix. `dexp` should no longer be requried

nextflow/configs/a549/virtual_stain.yml
```diff
@@ -0,0 +1,58 @@
...
+    test_time_augmentations: true
+    tta_type: 'median'
+
+data:
+  class_path: viscy_data.hcs.HCSDataModule
+  init_args:
+    source_channel: Phase3D
+    target_channel:
+      - "nuclei_prediction"
+      - "membrane_prediction"
```
ieivanov, line 35: Does this run successfully? This way renaming these channels is no longer needed, correct?

ieivanov:
Very exciting!

ieivanov:
## `uv sync` resolved — the root cause was a stale `uv.lock` entry, not `uv sync`

Follow-up on the environment guidance in this PR and on #308. Short version: the churn was a real bug, it had a one-line fix, and the per-task `uv run` wrapper is now gone entirely.

### What was actually wrong

`uv.lock` recorded `viscy-transforms==0.0.0.post215.dev0+4b62365`, but the wheel built from that **same pinned commit** declares `0.1.0a0`:

```
DEBUG Installed version does not match resolved version for viscy-transforms:
      0.1.0a0 vs. 0.0.0.post215.dev0+4b62365
DEBUG Requirement installed, but not fresh: viscy-transforms==0.1.0a0
```

uv therefore never saw the install as fresh and reinstalled it on every invocation — exactly the non-converging `Uninstalled 1 package / Installed 1 package` reported in #308. It is the only one of the five VisCy subpackages whose lock version disagrees with its built version.

Fix: `uv lock --refresh-package viscy-transforms` → `Updated viscy-transforms v0.0.0.post215.dev0+4b62365 (4b62365c) -> v0.1.0a0 (4b62365c)`. Same commit, one line of `uv.lock`.

After it, `uv sync --dry-run` reports **"Would make no changes"** and three consecutive `uv run`s show zero churn. `--no-sync` is not needed.

Worth watching for again: this recurs whenever a git-sourced dependency's declared version changes without its pinned rev changing. Diagnose with:

```bash
uv sync --dry-run -v 2>&1 | grep -i "does not match resolved version"
```

### The pipeline no longer wraps tasks at all

Rather than fixing the wrapper, this PR removes it: no `--biahub_project`, no `biahub_cmd()`, no `stain_cmd()`. The environment is resolved **once** before launch and inherited:

```bash
uv sync --project <BIAHUB>
source <BIAHUB>/.venv/bin/activate
nextflow run <BIAHUB>/nextflow/mantis-v2.nf ...
```

`sbatch` exports the submit environment (`--export=ALL` is the default) and the venv is on shared storage, so compute nodes resolve the same absolute paths. Verified directly on `cpu-c-2`: `VIRTUAL_ENV` and `PATH` propagate, `biahub` and `viscy` resolve to `.venv/bin`, and `cytoland` imports. `mantis-v2.nf` now calls `check_environment()`, which aborts at launch with an actionable message instead of failing N tasks with `command not found`.

`--extra stain` turned out to be unnecessary: the `dev` dependency group depends on `biahub[all]`, so a plain `uv sync` already installs `cytoland`.

### The 27-package hazard is checkout-specific

On my checkout, `uv sync --all-extras --dry-run` wants to remove **4** packages — `nd2`, `ome-types`, `xsdata`, `resource-backed-dask-array` — which are the `cellpose[all]` closure. Nothing in biahub imports them (`cellpose` is installed without `[all]`). A real `uv sync` took ~1s, hit no Lustre `ENOTEMPTY`, and left every critical import working.

So `caveats.md` §11 no longer forbids `uv sync`. It now tells you to *measure* the checkout in front of you and read the list before syncing: unused leaves → sync; anything load-bearing → stop and ask.

@tayllatheodoro — the 27 out-of-band packages (`cupy-cuda12x`, `tracksdata`, `stitch`, `dexp`, `dask-cuda`, `ilpy`, `pyscipopt`, `polars`, `pims` + closure) are specific to `/hpc/mydata/taylla.theodoro/repo/biahub`, and they are what made `uv sync` destructive there — not `uv sync` itself. Could you get that venv to a state where sync is a no-op? Current delta:

```bash
uv sync --all-extras --dry-run 2>&1 | grep -E "^ *- "
```

Two ways out, whichever fits how those packages are used:

1. **Declare them** — if the pipeline genuinely needs any of them (`cupy-cuda12x` for GPU paths, `tracksdata`/`ilpy`/`pyscipopt` for tracking), add them to `pyproject.toml` behind an extra and re-lock, so `uv sync` preserves them everywhere and everyone gets the same environment. This is the better outcome if they are load-bearing for real runs.
2. **Recreate from the lockfile** — if they were exploratory, `rm -rf .venv && uv sync` gives a clean, reproducible env matching this repo.

Until one of those happens, `uv sync` will keep pruning them on that checkout, and the new §11 will correctly tell you to stop. Happy to help with either — mainly I'd like to know whether any of those 27 are needed for a real pipeline run, since that decides between (1) and (2).

### Verification

End-to-end on `2026_03_24_A549_SEC61_ZIKV` (`--max_positions 1`, slurm profile): **16 tasks, 0 failures, 0 retries, ~21 min**, all six steps. `5-assemble` → `(10, 6, 86, 1600, 1370)` float32 with `['BF - Oblique', 'GFP EX488 EM525-45', 'mCherry EX561 EM600-37', 'Phase3D', 'nuclei_prediction', 'membrane_prediction']`; `4-track` → `nuclei_prediction_labels`. **Zero `uv run` in any `.command.sh` and zero env mutation in any task log.** 126 tests pass; ruff and pre-commit clean.

### Unrelated bug found while testing: `--max_positions` never worked

The smoke-test flag this PR advertises aborted every run that used it:

```
[ERROR] Missing process or function take([DataflowStream[?], 1])
```

A param passed on the command line arrives as a **String**, and `take` has no String overload, so Nextflow fell back to resolving `take` as a process. Only the CLI path was affected — the config default (`0`, typed) never reaches `take` — which is why production runs were fine and only the smoke test broke. Coerced with `as int` in `collect_positions`, and likewise for `queueSize`. New `caveats.md` §13 documents the trap for anyone adding a numeric param.

---

Closes #308 when this merges.

ieivanov:
Working through the review comments. Status on each:

### Done in this branch

**Store selection** (`SKILL.md:61`) — dropped the `du -sh`. Same stem with several indices now resolves to the highest index (`ls -1d ... | sort -V | tail -1`), with the reasoning recorded: `du -sh` walks millions of chunk files on Lustre *and* does not actually distinguish a large aborted acquisition from a small complete one. Different *stems* still stop and ask — that rule caught the `2026_08_05_dynatrack_2dpf` / `..._2df_*` typo, so it stays.

**Prefer configs on main** (`SKILL.md:96`) — precedence inverted. `<BIAHUB>/nextflow/configs/<family>/` is now the baseline; a previous run's `configs/` is a fallback and is explicitly framed as evidence, not a starting point ("unreviewed snapshots that drift with the schema"). Added the PR-back rule you asked for: a change that is *not* dataset-specific goes back to `nextflow/configs/` as a PR stating which dataset exposed it and what failed, with a stated test for the distinction — would the next dataset in this family want the same value? Per-dataset values (pixel size, scan step, one-off checkpoints) stay in the run directory. The plan step now reports the config commit SHA and any intended upstream PR.

**Torn-shard recovery** — restructured to resume-first. Step 1 is a plain `-resume`, since iohub#455 replaces a torn shard rather than reading it back and resumes per unit. Step 2, only if it fails *identically* again, is to **write up a repair proposal** — exact error, paths, `du -sh` of each, what is lost, what gets recomputed — and hand it to you or to the `job-io-error-repair` agent. **"Never delete zarr data from this skill"** is now stated in both `SKILL.md` §10 and `recovery.md`, and the decision table rows point at resume-first instead of straight at the repair agent. Also added: if a repair *is* still needed on a current checkout, say so explicitly, because that means per-unit resume did not cover the case and is worth knowing upstream.

**Invocation** — now `/reconstruct-with-nextflow <DATASET_NAME>`.

**`cd` into the output directory** — kept, and documented why, since the answer was not what the TODO assumed. The work dir does **not** depend on cwd: `nextflow.config` pins `workDir = "${params.output}/nextflow/work"` regardless of where you launch, overridable with `-work-dir` for scratch. What cwd *does* control is where `.nextflow.log` and the `.nextflow/` **resume cache** land — so relaunching from a different directory silently loses `-resume` and recomputes everything. That is the real reason to keep the `cd`.

### Filed

**Shard auto-sizing** (`caveats.md:237`) — https://github.com/czbiohub-sf/iohub/issues/458, covering both forms you named: a byte target (~2 GB) and a spatial-extent target, with the constraint that channels cannot be sharded folded in, plus a note that T is the cheapest axis to grow and the most damaging when a shard tears. `caveats.md` §7 links it and says the section goes away once it lands.

### Resolved elsewhere

**`list_positions` / 0-convert** — we will fix this at acquisition time by writing zebrafish acquisitions as HCS plates, rather than teaching the pipeline a second layout. Noted in `caveats.md` §1 as "being fixed at the source", with 0-convert still documented for existing data.

For the record, since it came up: `list_positions` *alone* already works on a raw zebrafish store with the pinned iohub —

```
$ biahub nf list-positions -i .../2026_08_05_dynatrack_2df_1.ome.zarr
1Pos0  1Pos1  1Pos2  1Pos3  2Pos0  2Pos1  2Pos2  2Pos3
```

— but that is not sufficient to drop the shim: those keys are one level deep, while every `init_*` step globs `-i <store>/*/*/*`, track configs use `fov: "*/*/*"`, and concatenate builds a plate. Fixing the acquisition is the cheaper path.

**#309** — root cause documented at https://github.com/czbiohub-sf/biahub/issues/309#issuecomment-5260761794. Short version: cellpose is not involved, `ultrack[gpu]` does not exist, and the trigger is `cupy` installed without `cucim` (ultrack picks `xp = cupy` on cupy alone but gets its accelerated `skimage` from `cucim`, silently falling back to CPU skimage). Recommendation is to add neither, since tracking takes 56.6s per position against virtual-stain's 8m45s. **No changes made in this PR for it.**

**`uv sync`** (`SKILL.md:168`, `caveats.md:252`) — done earlier, see https://github.com/czbiohub-sf/biahub/pull/305#issuecomment-5258295436. `dexp` confirmed not a biahub dependency; it is out-of-band on the other checkout.

### Still open, for you

- **#306** (optional `concatenate`/`track` configs) — not addressed here; the zebrafish placeholder `track.yml` and its caveat stay until it lands.
- **CLAUDE.md: update this skill with every PR merged to main** — I would rather agree the wording with you than guess at a repo-wide convention.
- **`module load cuda/12.8.0_570.86.10`** — not yet tested. torch wheels bundle their own CUDA runtime, so I expect it is unnecessary; I will confirm against a GPU step before removing the line.
- **`target_channel: nuclei_prediction`** — answered inline: yes it runs, verified end-to-end, but renaming is not fully obsolete.

nextflow/configs/a549/virtual_stain.yml
ieivanov: Yes — verified end-to-end today on `2026_03_24_A549_SEC61_ZIKV` (1 position, full six-step run, 0 failures). The assembled plate came out with the suffixed names directly from virtual-stain, no rename step involved:

```
5-assemble  channels=['BF - Oblique', 'GFP EX488 EM525-45', 'mCherry EX561 EM600-37',
                      'Phase3D', 'nuclei_prediction', 'membrane_prediction']
            shape=(10, 6, 86, 1600, 1370) float32
```

And `4-track` then resolved `target_channel: nuclei_prediction` against that plate and wrote `nuclei_prediction_labels`, which is the case that used to need the rename to have happened first.

**But renaming is not fully obsolete** — it is only the two VS channels that no longer need it. The same store still shows `BF - Oblique` (should be `BF`) and the acquired fluorescence channels with no `raw ` prefix, both of which are #291's convention and still come from `rename_channels.py` after the pipeline.

So the ordering hazard is gone for the channels track depends on, which was the dangerous part, but the rename step itself still has work to do until #291 moves it into `concatenate`.

.claude/skills/reconstruct-with-nextflow/templates/run_mantis_v2.sh
ieivanov: Tested, and you were right to question it — **the line is gone** (178fb3a).

It is not needed, and it was actively a mismatch. The torch wheel bundles its own CUDA runtime:

```
torch 2.13.0+cu130 | bundled CUDA 13.0
```

so the GPU steps only need the driver, which the GPU nodes already have. `module load cuda/12.8.0_570.86.10` put **12.8** on `LD_LIBRARY_PATH` ahead of torch's bundled 13.0.

Verified on an L40S (driver 580.126.20) with no cuda module loaded at all:

```
  (confirmed: no cuda module)
NVIDIA L40S, 580.126.20
torch 2.13.0+cu130 | bundled CUDA 13.0
cuda available: True
device: NVIDIA L40S
torch GPU matmul OK: (2048, 2048) True
cellpose gpu flag: True
cellpose GPU eval OK: (256, 256) uint16
```

That covers both GPU consumers — virtual-stain (torch) and track (cellpose, also torch). So to your question "is it that tracking runs on CPU without it": no, cellpose still reports `gpu=True` and runs on the GPU.

I left a comment in the script recording the reasoning so it does not get re-added. `module load nextflow` and `module load uv` stay — those are real.

.claude/skills/reconstruct-with-nextflow/SKILL.md
ieivanov: Done in a198d85 — the `du -sh` is gone.

Same stem with several indices now takes the highest index, as you suggested:

```bash
ls -1d /hpc/instruments/cm.mantis/<DATASET>/*.ome.zarr | sort -V | tail -1
```

I also recorded the second reason not to use `du -sh`, which I think is the stronger one: it does not actually answer the question. A large *aborted* acquisition outweighs a small complete one, so size cannot distinguish them even after you have paid the several minutes to walk the chunk files. The index can.

Kept the stop-and-ask rule for differing **stems**, since that is a genuinely ambiguous case rather than a resolvable one — it is what caught `2026_08_05_dynatrack_2dpf` containing `2026_08_05_dynatrack_2df_*.ome.zarr` (missing `p`). The `_epi`/`_prescan`/`_fov_debug` variants fall under the same rule: different acquisitions, not indices of one.

.claude/skills/reconstruct-with-nextflow/SKILL.md
ieivanov: Done in a198d85 — precedence inverted.

`<BIAHUB>/nextflow/configs/{a549,zebrafish}/` is now the baseline. A previous run's `configs/` is explicitly demoted to a fallback and labelled for what it is: an unreviewed snapshot that drifts with the schema and carries dataset-specific edits nobody meant to generalize.

Also added the PR-back rule from your notes comment, with a test for when it applies — **would the next dataset in this family want the same value?**

- Yes (schema migration, corrected default, plain wrong value) → PR against `nextflow/configs/<family>/`, stating which dataset exposed it and what failed.
- No (pixel size, scan step, a one-off checkpoint) → stays in the run directory.

The plan step now reports the config source commit and any change headed upstream, so a fix cannot quietly stay in a run directory where the next dataset misses it.

One related tightening: step 5 now cross-references the branch check in step 1b, because templates are only as current as the checkout — copying from a checkout behind `main` copies stale templates. It tells you to check `git log --oneline HEAD..origin/main -- nextflow/configs`.

.claude/skills/reconstruct-with-nextflow/templates/run_mantis_v2.sh
ieivanov: Follow-up — verified against the **real `biahub track`**, not just a synthetic cellpose call, since "does tracking fall back to CPU without it" deserved a proper test.

A/B on one position of an assembled A549 plate (T=10, 1600×1370), identical config and commit, only the module differing:

| variant | result | elapsed | cellpose per frame | objects |
|---|---|---|---|---|
| **no** cuda module | COMPLETED rc=0 | 45s | `mean=7.5, min=0, max=12` | 13 |
| `module load cuda/12.8.0_570.86.10` | COMPLETED rc=0 | 68s | `mean=7.5, min=0, max=12` | 13 |

Same shape `(10, 1, 1, 1600, 1370)`, same `uint32`, same channel `nuclei_labels`, same max label id. With the module, `LD_LIBRARY_PATH` did pick up `/hpc/apps/x86_64/cuda/12.8.0_570.86.10/lib64` — it just made no difference to the outcome.

**Tracking does not fall back to CPU without it.** Cellpose reports `gpu=True` and segments on the GPU either way; torch uses its bundled CUDA 13.0 runtime and only needs the driver.

One thing the A/B turned up that is worth recording separately: the label arrays were **not** bitwise identical between the two variants (1.80% of voxels). I ran the same variant twice as a control, and it differed from itself by **1.06%** — so the pipeline is simply not bitwise reproducible, and the between-variant difference is the same order as the within-variant noise. Sources are cellpose GPU inference, ultrack's `n_workers=16` watershed with `random_seed='frame'`, and ILP solution ties.

I would not have caught that without the control run, and it would have looked like evidence against removing the module. Added as `caveats.md` §14, with the practical consequence: do not diff label arrays to judge whether a change was harmful — 1–2% is the noise floor, compare object counts instead. Deskew/reconstruct/assemble are unaffected and remain deterministic.

.claude/skills/reconstruct-with-nextflow/references/caveats.md
ieivanov: Filed as https://github.com/czbiohub-sf/iohub/issues/458, covering both forms you named — a byte target (~2 GB) and a spatial-extent target — with the caller expressing intent rather than computing a chunk multiple.

I folded in the constraint you mentioned: **channels cannot currently be sharded**, so an auto-sizer must pin the C entry at 1. That constraint is invisible in today's `shards_ratio` API — nothing stops you writing a C multiplier > 1 — so encoding it removes a footgun as well.

Also noted in the issue that T is special: growing T is the cheapest way to hit a byte target and the most damaging when a shard tears (a shard spanning 10 timepoints loses all 10), so the byte-target form should prefer spatial axes before T or make T growth explicit.

`caveats.md` §7 links the issue and says the section goes away once it lands. Resolving here; the work is tracked upstream.

ieivanov:
@tayllatheodoro — a request on the acquisition side, which would let us delete a whole class of workaround.

## Please switch zebrafish/dynatrack acquisitions to HCS plate format

The pipeline fans out over HCS plate positions, so a flat `bioformats2raw.layout: 3` store needs a `0-convert` symlink plate built before it can run at all. That shim is the single biggest source of per-dataset manual work in the reconstruct skill, and it exists purely because of how the acquisition is specified — not because of anything about the data.

**The fix is in the acquisition config: add `plate_row` and `plate_col` to each entry in `stage_positions`.**

### Current zebrafish — flat, no plate keys

`/hpc/instruments/cm.mantis/2026_08_05_dynatrack_2dpf/config.yaml`:

```yaml
stage_positions:
  - x: -5953.17
    y: 22522.82
    name: 1Pos0
    properties:
      - ['ZDrive', 'Position', 4643.36]
  - x: -6318.399
    y: 22564.879
    name: 1Pos1
    properties:
      - ['ZDrive', 'Position', 4645.46]
```

This produces a store with `1Pos0`, `1Pos1`, … at the root and no `attributes.ome.plate`.

### A549 — already an HCS plate

`/hpc/instruments/cm.mantis/2026_07_14_A549_MAP4_ZIKV/config.yaml`:

```yaml
stage_positions:
  - name: "000000"
    plate_row: 1
    plate_col: 1
    x: -4955.37
    y: 14204.75
    properties:
      - ['ZDrive', 'Position', 4546.92]
  - name: "001000"
    plate_row: 1
    plate_col: 1
    x: -4501.900000000001
    y: 12984.2
    ...
```

Those two extra keys are the whole difference. That config has 66 positions across 6 distinct `(plate_row, plate_col)` pairs at 11 FOVs each, and the resulting store is a proper plate with row/column groups (`B/`, `C/`, …) and fields beneath them — which the pipeline consumes directly, no conversion.

Note positions sharing a `(plate_row, plate_col)` become multiple FOVs in one well, so for zebrafish you can group however is biologically meaningful — one well per fish, per dish, per condition — as long as every position carries the two keys.

### Why this is worth doing at the source

- **It removes the shim entirely.** No `0-convert`, no `build_plate*.py` script per dataset, no symlink tree to verify before launching.
- **The shim has its own failure mode.** The existing build scripts carry a recovery fallback for positions whose group `zarr.json` was left empty by an interrupted instrument write — that complexity disappears too.
- **Half-fixing it downstream does not work.** Worth knowing: the pinned iohub *does* now read flat stores, so `biahub nf list-positions` already enumerates them:

  ```
  $ biahub nf list-positions -i .../2026_08_05_dynatrack_2df_1.ome.zarr
  1Pos0  1Pos1  1Pos2  1Pos3  2Pos0  2Pos1  2Pos2  2Pos3
  ```

  But that is not enough to skip conversion: those keys are one level deep, while every `init_*` step globs `-i <store>/*/*/*`, track configs use `fov: "*/*/*"`, and `concatenate` builds a plate. Teaching six pipeline steps two layouts is strictly more work than adding two keys to the acquisition config.

Existing data still needs `0-convert`, so the shim stays documented for it — but everything acquired after this change would skip it. Is there a reason the zebrafish acquisitions were set up flat that I am missing, or is this just how it started?

ieivanov:
## Production validation: full-plate run on this branch, 341/341 tasks

`2026_07_14_A549_MAP4_ZIKV_rerun_2` ran the whole pipeline on this branch with the new run script — `uv sync` + activate, no `--biahub_project`, bare `biahub`/`viscy` — on 66 positions × T=67, ~10h wall, **≈37.5 TB** of output.

```
Workflow completed > succeededCount=341; failedCount=0; retriesCount=0; cachedCount=0
                     peakRunning=30; peakCpus=480; peakMemory=9.8 TB
> Execution complete -- Goodbye
```

| step | tasks | max realtime |
|---|---|---|
| flat_field | 66 | 9m 23s |
| deskew | 66 | 9m 58s |
| apply_inv_tf | 66 | 9m 8s |
| virtual_stain_preprocess | 1 | 17m 2s |
| virtual_stain | 66 | 59m 57s |
| run_concatenate | 1 | **4h 41m** |
| track | 66 | 5m 52s |
| init/singleton steps | 8 | ≤48s |

Sizes: 14 TB `0-flatfield`, 4.9 TB `1-deskew`, 2.5 TB `2-reconstruct`, 4.1 TB `3-virtual-stain`, 489 MB `4-track`, 12 TB `5-assemble`.

`5-assemble` came out `(67, 6, 86, 1664, 1193)` float32 with #291-normalized channels (`BF`, `raw GFP EX488 EM525-45`, `raw mCherry EX561 EM600-37`, `Phase3D`, `nuclei_prediction`, `membrane_prediction`); `4-track` has `nuclei_prediction_labels` for all 66 positions.

### This is the evidence for the environment change

The point of removing the per-task `uv run` wrapper was that up to `maxForks` tasks were re-materializing one shared `site-packages`. This run exercised that at **30 concurrent tasks**:

- **0 of 341** `.command.sh` files contain `uv run`
- **0** task logs show `Uninstalled N package` / `Installed N package`

So #308's contention scenario is now empirically clear, not just argued.

## `module load cuda` — verified unnecessary for **both** GPU consumers

The run above used `module load cuda/12.8.0_570.86.10`, which made it a clean reference to A/B against. Both GPU steps were tested with the module removed.

**virtual-stain** — same position (`B/2/000000`) off the same `2-reconstruct`, same config and checkpoint, no cuda module, on a different GPU node:

```
  no cuda module loaded
NVIDIA L40S, 580.126.20
torch 2.13.0+cu130 bundled CUDA 13.0 available True
### rc=0 elapsed=2797s        (reference run's max was 59m 57s)
```

Compared against that run's own `3-virtual-stain` output for the same position:

```
  t=0   full volume (2, 86, 1664, 1193): differing voxels = 0 / 341,446,144
  t=33  full volume (2, 86, 1664, 1193): differing voxels = 0 / 341,446,144
  t=66  full volume (2, 86, 1664, 1193): differing voxels = 0 / 341,446,144

TOTAL over 3 full timepoints: 0 differing of 1,024,338,432 voxels
BITWISE IDENTICAL: True
```

**track** — separate A/B on an assembled plate, reported in the inline thread: both variants COMPLETED with identical cellpose counts (`mean=7.5, min=0, max=12`), 13 objects, same shape/dtype/channel. Cellpose reports `gpu=True` either way, so tracking does not silently fall back to CPU.

The module was not merely redundant — it put CUDA **12.8** on `LD_LIBRARY_PATH` ahead of the 13.0 runtime bundled in the torch wheel. Removed in 178fb3a with the reasoning recorded in the script.

### Two findings worth keeping

**Tracking is not bitwise reproducible; virtual-stain is.** The track A/B showed a 1.80% voxel difference between variants, which read as evidence the module mattered. A control — the same variant run twice — differed from itself by **1.06%**, so that was run-to-run noise: cellpose GPU inference, ultrack's `n_workers=16` watershed with `random_seed='frame'`, and ILP solution ties. Virtual-stain, by contrast, is bitwise reproducible even across GPU nodes and CUDA runtimes, as above. Both recorded in `caveats.md` §14, with the practical rule: a 1–2% tracking difference is the noise floor and object counts are the thing to compare, whereas *any* virtual-stain difference is real and worth investigating.

**Scope of the cuda claim.** Two GPU consumers, one dataset family, L40S nodes. It shows the module is not load-bearing and is a version mismatch; it is not a sweep across every node type in the cluster. If a GPU step ever fails with a CUDA error on hardware not tested here, this is the commit to look at first.

ieivanov:
## Decision: tracking stays on CPU — no `cupy`/`cucim` in the environment

ultrack has an optional GPU path, and its own `pyproject.toml` suggests it:

```toml
[tool.pixi.feature.cuda.dependencies]
cupy = "*"
cucim = "*"
pytorch-cuda = "*"
```

(That is a conda/pixi feature. `pytorch-cuda` has no pip equivalent — our torch is already a CUDA build, `2.13.0+cu130` — and ultrack's `system-requirements` target **CUDA 12** while we are on 13.)

Measured rather than assumed, in an **isolated venv** built from this repo's lockfile so the production `.venv` was never modified (verified before and after: no cupy present, `uv sync --dry-run` reports no changes). Same position, same config, same node type:

| config | ultrack `xp` | cucim | result | objects | elapsed |
|---|---|---|---|---|---|
| **no cupy** (current baseline) | numpy | absent | OK | 13 | **48s** |
| cupy + cucim | cupy | 26.08.00 | OK | 13 | **67s** |
| cupy + cucim (repeat) | cupy | 26.08.00 | OK | 13 | **46s** |
| cupy, no cucim | cupy | absent | **FAILS** rc=1 | **0** | 37s |

**No speedup.** 46–67s against a 48s baseline is noise, and tracking is ~1 min per position against virtual-stain's ~46 min — it is not the bottleneck. Identical object counts throughout. So the GPU stack costs two extra CUDA-pinned dependencies and buys nothing measurable, while introducing the failure in row 4.

Not adopting it. Recorded as a comment above the `track` extra in `pyproject.toml`, including the conditions if it is ever revisited: add **both** packages (never one alone), pin the CUDA major to the torch wheel (`cu130` → `cupy-cuda13x` + `cucim-cu13`), and benchmark on a full plate first.

## #309 reproduced, root cause confirmed

Installing `cupy` **without** `cucim` reproduced it exactly:

```
  ultrack xp module: cupy | cupy active: True
  cucim: ABSENT
Cellpose cells per frame: mean=7.5, min=0, max=12      <- cellpose SUCCEEDS
Tracking with cellpose labels...
  File ".../ultrack/utils/edge.py", line 85, in labels_to_contours
    contours_frames += segm.find_boundaries(lb_frame, mode="outer")
  File ".../skimage/segmentation/boundaries.py", line 164, in find_boundaries
TypeError: Implicit conversion to a NumPy array is not allowed.
```

ultrack sets `xp = cupy` on cupy availability alone, but sources its accelerated `skimage` from `cucim` and falls back silently to CPU `skimage` when that is missing. So `labels_to_contours` builds a cupy array and hands it to a CPU function. Confirms the analysis in #309 and that the trigger was an out-of-band `cupy` on one checkout, not a biahub bug.

Two details worth recording:

- **The failed run still wrote an output store, with 0 labels.** A masked failure here yields silently empty tracking rather than an obvious crash.
- **`< /dev/null` makes the `(Pdb)` prompt exit `rc=1` in 37s** instead of hanging to the wall limit. Noted on #311.

## Segmentation is unaffected

Cellpose reported identical per-frame counts (`mean=7.5, min=0, max=12`) in **all four** configurations, including the one that failed downstream. Segmentation runs on torch and never touches cupy — consistent with the earlier finding that removing `module load cuda` did not push it to CPU either.

## Virtual-stain

Still running as I post this — the remaining question is only whether *having* cupy/cucim installed perturbs it. It runs on torch via cytoland, and the earlier no-cuda-module test already showed it bitwise reproducible (0 differing voxels in 1,024,338,432), so I expect no effect. I will follow up with the comparison either way rather than leave it implied.

ieivanov (approved):

ieivanov:
## Follow-up: virtual-stain with `cupy`+`cucim` installed

Completing the comparison promised above. Same position (`B/2/000000`), same config and checkpoint, no cuda module, but with `cupy-cuda13x` + `cucim-cu13` present in the environment. Reference is `rerun_2`'s own `3-virtual-stain` for that position.

**The predictions are numerically equivalent.**

| channel | corr | max abs diff | median abs diff | ref mean / std | new mean / std |
|---|---|---|---|---|---|
| `nuclei_prediction` | 1.000000 | 9.3e-04 | 5.6e-06 | +2.1935 / 6.1161 | +2.1935 / 6.1161 |
| `membrane_prediction` | 1.000000 | 2.4e-04 | 6.6e-06 | +0.6144 / 0.7775 | +0.6144 / 0.7775 |

Identical mean, std, min and max to four decimals; `max abs diff` is 0.002% of the ~39 value range. So virtual-stain is unaffected in any way that matters, which is expected — it runs on torch and never touches cupy.

**One correction to something I posted earlier.** The output was *not* bitwise identical this time — 1,022,977,764 of 1,024,338,432 voxels differ, i.e. 99.87%. That number is meaningless on its own: the differences are float32 last-bit rounding, almost certainly from different GPU kernel selection once the extra `nvidia-*-cu13` libraries that come with cupy/cucim are on the path.

I previously reported virtual-stain as "bitwise reproducible" on the strength of a 0-differing-voxel result between the with- and without-cuda-module runs. That holds **within one installed library set**, but not across environments whose CUDA libraries differ. I have corrected `caveats.md` §14, which had generalised it into "any virtual-stain difference is real and worth investigating" — that would have sent someone chasing a 99% voxel diff that is actually noise. It now contrasts the two failure modes explicitly and says to judge virtual-stain on `corr` and `max abs diff`, not voxel counts, since tracking's differences are structural (different label ids) while virtual-stain's are last-bit.

**Possible slowdown, not confidently attributable.** 4093s (68m 13s) with the GPU stack installed, against two independent ~46 min measurements for the same position — 2749s inside the production run (sharing a node with 30 concurrent tasks) and 2797s alone on a node without cupy. Suggestive, but n=1 on a different node, so I would not claim a 46% penalty from one sample. It certainly does not make virtual-stain faster.

## Net effect on the decision

Nothing changes: still no `cupy`/`cucim`.

| step | effect of installing cupy+cucim |
|---|---|
| segmentation (cellpose, torch) | none — identical counts in all four configs |
| tracking (ultrack) | no speedup (46–67s vs 48s CPU baseline), identical objects; **breaks outright if cucim is absent** |
| virtual-stain (cytoland, torch) | numerically equivalent; possibly slower; two extra CUDA-pinned dependency trees |

Two packages, one hard failure mode when they drift apart, and no measured benefit anywhere.

PR #314, opened 2026-08-14
perf(flat-field): tile the median so it stays in cache (19x on Zen 2 nodes)

ieivanov:
## Problem

`flat_field_zyx` reduces with `np.median(zyx_data, axis=0)`. Axis 0 has a stride of `Y * X * itemsize` — **852 KB** for a mantis-v2 position — so the partition underneath `np.median` touches one cache line per element and runs at DRAM latency.

`np.partition` alone accounts for **134.35 s of the 134.66 s** kernel. The float64 divide is 0.61 s, and casting the median to float32 saves nothing — the partition *is* the kernel.

That cost lands very unevenly across Bruno. Measured on identical synthetic data, single-threaded, zero I/O, across 17 nodes:

| CPU | nodes | strided median | contiguous | streaming |
|---|---|---|---|---|
| EPYC 9354P / 9534 (Zen 4) | `cpu-h`, `cpu-i`, `cpu-g` | 2.05–2.09 s | 0.85–0.88 s | 7.8–9.4 GB/s |
| EPYC 7713 / 7763 / 7773X (Zen 3) | `gpu-b`, `cpu-e`, `gpu-d` | 2.32–2.57 s | 0.86–0.93 s | 8.3–9.4 GB/s |
| EPYC 7313P (Zen 3) | `gpu-e` | 2.30 s | 0.86 s | 9.5 GB/s |
| **EPYC 7742 (Zen 2)** | **`gpu-a-1..4`** | **12.59–12.96 s** | 1.09–1.14 s | 8.3–8.6 GB/s |
| **EPYC 7302P (Zen 2)** | **`gpu-sm01/02`** | **13.04–13.31 s** | 1.15–1.17 s | 7.1–7.7 GB/s |

The affected nodes are within 12% of everything else on streaming bandwidth and within 1.3x on a *contiguous* median — they are only slow at the **strided** reduction. All four `gpu-a` nodes agree within 3%, and a second unrelated Zen 2 model in a different chassis behaves the same, so this is not a degraded node.

In the `2026_08_11_A549_SEC61B_DENV` run this made flat-field spend **~139 CPU-s per (t, c) unit** on `gpu-a`/`gpu-sm` against **~12 CPU-s** on `cpu-h` — a 10x spread over 54 positions × 201 units, and the reason 8 tasks hit the 2h30 wall limit and had to retry.

Ruled out along the way: thread oversubscription (`cpu/wall = 1.00` in every phase; pinning `OMP/MKL/OPENBLAS/NUMEXPR/BLOSC` changed nothing), worker concurrency (+4.7% CPU/unit from 1→16 workers on `gpu-a`), write I/O (4.69 s vs 4.21 s per unit between the two classes), RAM pressure (zero `OUT_OF_MEMORY`; MaxRSS 57–107 GB against a 224 GB request), and dtype promotion.

## Fix

Copy a slim tile into a compact buffer before reducing, sized to fit the smallest per-CCX L3 in the cluster (16 MiB on Zen 2). Per-position median on real data:

| node | CPU | `np.median` | tiled |
|---|---|---|---|
| `gpu-a-3` | EPYC 7742 | 134.33 s | **5.52 s** |
| `gpu-sm02-19` | EPYC 7302P | 99.26 s | **5.72 s** |
| `cpu-e-1` | EPYC 7763 | 9.89 s | **5.18 s** |

The three classes converge to within 10% of each other — the 13.6x hardware spread collapses to 1.1x. End to end through `flat_field_zyx` on one real unit: **131.12 s → 6.79 s (19.3x), output bit-identical.**

Transposing to make the reduction axis contiguous was also tried and is far worse on the affected nodes (78.25 s on `gpu-a-3`) — the transpose copy is itself a strided read. Only tiling works.

## Correctness

Output is unchanged by construction: each 1-D lane is reduced from the same elements by the same algorithm, so tiling only changes the access pattern. Verified bit-identical (`np.array_equal`) against the current implementation on a real mantis position, and `tests/test_flat_field.py` asserts exact equality against `np.median` across dtypes (`uint16`/`float32`/`float64`), all three axes, even *and* odd reduction-axis lengths (the even case averages two middle values), degenerate axes, non-contiguous views, and forced tile widths from 1 byte upward.

One test is a performance guard rather than a correctness check: it asserts a real position does not silently collapse back to a single whole-array tile, which would stay correct but lose the 11.5x.

## Notes

- Wall-clock gains are concentrated where CPU was the binding constraint. On `cpu-h`-class nodes flat-field is already I/O-bound (12.3 CPU-s against ~206 s wall per unit), so those get little faster; the win is the CPU-hours and the `gpu-a` tasks that were timing out.
- `deskew` was checked for the same pattern and does **not** have it — Zen 2 `gpu-sm` nodes are the *fastest* class at deskew (6.5 min mean), so this is specific to flat-field's median.
- Incidental finding, not addressed here: SLURM tags both `gpu-a` and `gpu-b` as `amd_7742`, but `gpu-b` reports an EPYC 7713. `--constraint=amd_7742` does not select what it claims.

### DIFF
--- a/biahub/flat_field.py
+++ b/biahub/flat_field.py
@@ -33,6 +33,77 @@ from biahub.cli.utils import (
 )
 from biahub.settings import FlatFieldCorrectionSettings
 
+# Byte budget for one median tile. The reduction below runs at DRAM latency when
+# its working set spills out of cache, so the tile is sized to fit in the smallest
+# per-CCX L3 in the cluster (16 MiB on Zen 2), with headroom for the partition's
+# own scratch. See :func:`_median_tiled` for why this matters.
+_MEDIAN_TILE_BYTES = 8 * 1024**2
+
+
+def _median_tile_axis(data: np.ndarray, axis: int) -> int | None:
+    """Return the first axis that is not the reduction axis, or None for 1-D input."""
+    return next((a for a in range(data.ndim) if a != axis), None)
+
+
+def _median_tile_width(data: np.ndarray, axis: int, tile_bytes: int) -> int:
+    """Return how many indices along the tiled axis fit in ``tile_bytes``.
+
+    Never less than 1, never more than the axis length.
+    """
+    tile_axis = _median_tile_axis(data, axis)
+    bytes_per_step = data.itemsize * (data.size // data.shape[tile_axis])
+    width = tile_bytes // max(bytes_per_step, 1)
+    return int(min(max(width, 1), data.shape[tile_axis]))
+
+
+def _median_tiled(
+    data: np.ndarray, axis: int, tile_bytes: int = _MEDIAN_TILE_BYTES
+) -> np.ndarray:
+    """``np.median(data, axis=axis)``, evaluated over cache-resident tiles.
+
+    Output is identical to :func:`numpy.median` -- only the memory access pattern
+    changes. ``np.median`` partitions along ``axis``, whose stride in a ZYX volume
+    is ``Y * X * itemsize`` (852 KB for a mantis-v2 position), so every element
+    touched lands on its own cache line and the reduction runs at DRAM latency.
+    That cost is borne unevenly by the cluster: measured on one position, it is
+    11.5x worse on Zen 2 cores (EPYC 7742, 7302P) than on Zen 3 and newer, which
+    made flat-field spend ~139 CPU-s per (t, c) unit on ``gpu-a``/``gpu-sm`` nodes
+    against ~12 CPU-s on ``cpu-h``.
+
+    Copying a slim tile into a compact buffer first keeps the partition in cache.
+    Per-position medians measured before and after:
+
+    ==================  ===========  =========  ========
+    node                CPU          np.median  tiled
+    ==================  ===========  =========  ========
+    ``gpu-a-3``         EPYC 7742    134.33 s   5.52 s
+    ``gpu-sm02-19``     EPYC 7302P    99.26 s   5.72 s
+    ``cpu-e-1``         EPYC 7763      9.89 s   5.18 s
+    ==================  ===========  =========  ========
+
+    Transposing to make the reduction axis contiguous was also tried and is much
+    worse on the affected nodes (78.25 s on ``gpu-a-3``), because the transpose
+    copy is itself a strided read.
+    """
+    tile_axis = _median_tile_axis(data, axis)
+    if tile_axis is None:
+        return np.median(data, axis=axis)
+
+    tile = _median_tile_width(data, axis, tile_bytes)
+    tiles = []
+    for start in range(0, data.shape[tile_axis], tile):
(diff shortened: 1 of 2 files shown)


ieivanov:
## No-regression check on the fast nodes

The tiling adds an `ascontiguousarray` copy per tile, which is pure added work on the classes that were never slow — and flat-field runs 16 worker processes per position, so a change that wins single-threaded could still lose when 16 copies compete for bandwidth. Benchmarked both implementations against each other on a real mantis volume, loaded pre-fork so children inherit it copy-on-write and both arms see identical data:

| node | CPU | 1w old → new | 1w gain | 16w CPU-s/unit old → new | 16w gain | peak RSS old / new |
|---|---|---|---|---|---|---|
| `cpu-h-5` | EPYC 9354P (Zen 4) | 8.81s → **4.50s** | 1.96x | 11.04s → **5.29s** | **2.09x** | 4.9G / 4.9G |
| `cpu-i-1` | EPYC 9354P (Zen 4) | 9.23s → **4.52s** | 2.04x | 9.24s → **5.06s** | 1.83x | 4.9G / 4.9G |
| `cpu-g-1` | EPYC 9534 (Zen 4) | 9.20s → **4.69s** | 1.96x | 10.40s → **6.39s** | 1.63x | 4.9G / 4.9G |
| `cpu-e-1` | EPYC 7763 (Zen 3) | 10.67s → **5.82s** | 1.83x | 15.31s → **9.99s** | 1.53x | 4.9G / 4.9G |
| `gpu-a-2` | EPYC 7742 (Zen 2) | 136.30s → **6.27s** | 21.75x | 164.69s → **9.30s** | 17.71x | 4.9G / 4.9G |

**No class regresses.** Every `cpu-*` node gains 1.5–2.1x, at both single-process and production's 16-worker concurrency. Output verified bit-identical on every node before timing.

Three points from the data:

- **No memory cost.** Peak child RSS is 4.9 G in all ten arms. The 7.1 MB tile copy is invisible against the 3.6 GB float64 temporaries the kernel already allocates.
- **Less sensitive to concurrency, not more.** Going 1→16 workers raises CPU-s/unit in both arms from bandwidth contention (cpu-h-5: 8.79→11.04 old, 4.47→5.29 new), but the ratio holds or improves. A cache-resident tile moves less memory traffic, so it competes less.
- **Corroborates the production numbers.** `gpu-a-2` at 16 workers measures 164.69 CPU-s/unit for the current code, against the ~139 CPU-s/unit derived independently from `sacct` on the live 2026_08_11_A549_SEC61B_DENV run.

Benchmark: `bench_ab.py` in `/hpc/mydata/ivan.ivanov/ff_thread_test/` on Bruno. A `gpu-b` (EPYC 7713) arm was queued but the node never freed up; Zen 3 is represented by `cpu-e-1`.

Issue #199, opened 2026-02-14
Zarr v2 -> Zarr v3 migration across computational imaging team's frameworks


mattersoflight:
We have decided to upgrade our image processing and analysis pipelines from zarr v2 (ome-zarr-0.4) to zarr v3 (ome-zarr-0.5) format.
The key reason for the switch is that the support for sharding and small chunks in zarr v3 datasets will enable real-time inference for smart microscopy, efficient model training through parallel/distributed data loading, and efficient visualization. 

This issue is focused on identifying any disruption in processing that computational imaging team and collaborators may encounter due to the change in format, across `iohub`, `biahub`, `waveorder`, `viscy`, and `shrimpy`. We are centralizing the discussion in `biahub` repo.

`iohub` has been updated to support reading and writing with both versions: When we write the zarrs during the processing steps we just need to set the version flag in `iohub.ngff.open_ome_zarr` to `v0.5`.

We need to make a few key decisions:
1. Should the refactored CLI across the repos "silently" migrate from zarr v2 to zarr v3, or propagate the format? For silent migration, the CLI can read zarr v2/v3 data, and write only zarr v3 data. This is currently preferred to migrate to a better format and reduce data duplication. If the conversion can break existing analysis workflows and the fix is not straightforward, we can propagate the format. 

2. Which CLI need to be updated in what order so that silent migration doesn't break existing analysis workflows? We need key users's inputs to decide if migration to zarr v3 will break their data analysis pipelines. 

 @Soorya19Pradeep, @tayllatheodoro, @gav-sturm can you please add a sub-issue to clarify which CLI are you currently dependent on that could break with migration to zarr v3? We can use the sub-issue to resolve the problems.

---
## Current zarr v2 --> v3 transition plan
 - [x] https://github.com/czbiohub-sf/iohub/issues/401
 - [ ] https://github.com/czbiohub-sf/biahub/pull/215
 - [ ] https://github.com/czbiohub-sf/biahub/issues/240
 - [x] https://github.com/mehta-lab/waveorder/issues/540
 - [x] https://github.com/mehta-lab/waveorder/issues/550
 - [x] https://github.com/mehta-lab/VisCy/pull/407
 - [ ] https://github.com/czbiohub-sf/iohub/issues/402

---
## Test plan
- [ ] mantis v2 reconstruction should proceed without errors. Intermediate reconstructions should be written in zarr v3 with proper chunking, sharding, and compression
[...]

PR #322, opened 2026-08-19
feat(qc): imaging-qc as a direct dependency, and QC of the assembled store in mantis-v2

ieivanov:
Supersedes #298 — this branch is #298 plus four commits, so close that one if this is the PR you want to review.

## What this does

QC runs as part of a reconstruction. `mantis-v2.nf` gains `--qc_config`; pointed at a stage config it QCs the **assembled** plate, the one store carrying every channel the pipeline produced. Running QC on individual step stores stays available through `qc-standalone.nf`, which now emits one report per store.

**Blocked on [imaging-qc-pipeline#199](https://github.com/czbiohub-sf/imaging-qc-pipeline/pull/199).** `pyproject.toml` pins `imaging_qc_pipeline` at that PR's branch tip, because imaging-qc pinned `iohub==0.3.2` while biahub needs `iohub>=0.3.11` for the gapped-shard-write fix (iohub#460) — two exact pins in one environment have no solution. #199 relaxes it to a floor. Merge that, then repoint the rev at a main commit; the `[tool.uv.sources]` comment says so.

## The environment contract

`qc_cmd()` prefixed every QC task with `uv run --from 'imaging-qc-pipeline @ git+…'`, so up to `maxForks` tasks re-resolved an environment and re-fetched a git dependency concurrently against one shared site-packages. QC was the last holdout after `biahub_cmd()` went away; it now follows the same contract — resolve once, call `imaging-qc` bare:

```bash
uv sync --extra qc
source .venv/bin/activate
```

The `qc` extra is deliberately **not** in `all`: imaging-qc-pipeline is a private repo with no PyPI release, and CI's `uv sync --group dev` resolves `all` with no credentials for it. `check_environment()` now takes the tools the caller actually runs, so mantis-v2 isn't held to an `imaging-qc` it never calls, and a missing CLI fails at launch instead of N tasks in.

Two latent breaks fell out: `qc_processes.nf` still imported `biahub_cmd` from `common.nf`, which no longer defines it, and the standalone usage comment named a file (`nextflow/qc.nf`) that doesn't exist.

## Where QC sits in the pipeline

It gates on `assemble_done` — the same signal `track_wf` waits on — so QC runs **concurrently with tracking** and never extends the critical path.

A QC verdict cannot fail the run: `imaging-qc gate` exits 0 whether positions pass or fail (measured, with `fail=4`), recording the verdict in the store's `tables/qc/` tables and a `QC_SUMMARY` line. Only a broken config or a genuine compute error exits non-zero, which is what should stop a pipeline.

## Fan-out, and where the parallelism actually is

- **`qc_chunk_size` 10 → 0**: one compute task per position. imaging-qc dispatches a position's `(T, C)` units through a thread pool sized from `max_concurrent`, which defaults to the task's CPU affinity, so a position is already parallel inside its node. Chunking T only multiplies jobs — 100 positions × T=100 is **100 tasks at 0, 1000 at 10**.
- **`compute_step` gains `cpus`** (`qc_cpus`, default 8). It had none, so that thread pool saw one CPU and walked a position's timepoints serially.

## The report

One report per store, so the report-spec generator is gone. A spec exists to put N stores in one multi-tab report; with one store per report, `imaging-qc report <store> <dir>` covers it — which is what imaging-qc's own `qc_report.nf` does. Deleted: `biahub/qc.py`, its 7 tests, the `generate-report-spec` command, the `generate_report_spec` process, and with them the `qc_dir` resolution problem a generated spec created.

`report` is handed a config **directory**, not the config file. imaging-qc's report verb does not compose Hydra `defaults:` from a single file, so a config inheriting its `report:` block from `base.yaml` loses every metric plot and renders only the outlier heatmap — at exit 0, silently. Measured: file → 1 plot, directory → 5. `plan-stage` and `compute` do compose from the same file, so this looks like an upstream inconsistency worth an issue there.

## Config migration

Pinning imaging-qc exposed that the QC configs were written against an older one. Four keys it refuses or ignores, all found by running the configs rather than reading them:

| Key | Fate |
| --- | --- |
| `bleach_rate` (+ its `temporal` group, gate, report entries) | not a metric any more; replaced by nothing |
| `blank.gate.aggregation` | refused — imaging-qc's changelog names this file by path |
| `granularity: z` | retired for `slice`, same scope |
| `report.mode` | not a `ReportConfig` key; warned and ignored |

No threshold or metric set changes; gates evaluate as before.

## Verification

One Python 3.13 environment (`uv sync --extra qc`), both CLIs on PATH sharing iohub 0.3.11, plotly present for the default interactive report. Then the full chain against synthetic stores using the exact commands these processes emit — `plan-stage` → `estimate-resources` → `compute` → `consolidate --wave-id` → `consolidate` → `gate` → `report`:

- 4 positions × 4 channels × T=6 × Z=5: plan at chunk 0 gives one item per position with no chunk args; 4 computes and both finalizes clean; `gate` reports `positions=4`; report renders all 5 plots into the `<store>_report` sibling.
- Stage 5 gates fire per channel off the `{channel: value}` SNR floors (`low_signal_to_noise_raw GFP EX488 EM525-45`); stage 6 cell-count passes on a 2-instance frame against `hard_floor: 1`.
- Consolidated table grain confirmed: long format keyed `(position, T, C, Z, metric_name)` — slice metrics 480 rows each (4×4×6×5), `intensity_stats` keys 96 each (Z collapsed).

`ruff` clean. `tests/test_cli` passes apart from 3 pre-existing `napari` import failures that reproduce identically on a stashed tree.

**Not verified: Nextflow itself** — not installed on the box this was developed on, so the DSL edits are unexecuted. Every command inside the processes is verified, but the channel wiring (`combine` on `assemble_done.done`, the `if (qc_on)` block, the reshaped `qc_report_wf`) has only been read. A `--max_positions 1` smoke run is the next step, and the reason this is a draft.

## Before merge

- [ ] imaging-qc-pipeline#199 merges; repoint the `[tool.uv.sources]` rev at a main commit
- [ ] Smoke-run mantis-v2 with `--qc_config` on one position
- [ ] `uv-lock.yml`'s weekly `uv lock --upgrade` needs a credential (deploy key or PAT) for the private imaging-qc repo, or it will start failing
- [ ] The SNR `hard_floor` map still keys on DynaCell channel names; set floors against a real assembled store's channels or they silently apply to nothing
- [ ] `qc_cpus = 8` is a guess — tune against one real position's runtime
- [ ] Close #298 as superseded


ieivanov:
Closing in favour of **#298**, which is now the PR of record for this work.

The `qc-assembled-store` branch was merged into `qc-refactor-to-main` (a fast-forward, since it was a strict descendant), so #298 carries every commit that was here plus the follow-up work — no cherry-picks, no duplicated commits.

One thing that landed after this branch: the per-store report introduced here is replaced by a single multi-tab report (`${output}/qc/report`), matching the 2026-08-12 DynaCell run, with the spec written at launch by a Nextflow function rather than a `biahub` CLI call. See `2041f0b` on #298.

PR #330, opened 2026-09-08
build(deps): imaging-qc 0.5.0, and read Z stacks by max in the default QC configs

ieivanov:
Upgrades the QC dependency and, on the strength of what the new version measures, changes how the default configs read a Z stack.

## 1. Dependency: imaging-qc `0.4.0a2` -> `0.5.0` (`bd0b494`)

Verified the upgrade is metric-preserving before adopting it. Recomputed the whole `v3` A549 store on both versions and compared value by value: **5,882,868 / 5,882,868 identical**, and no position changed verdict.

What does change is intended, and is the reason for the bump:

- **Blank precedence.** A channel detected as blank no longer cascades a pile of derived flags; the dependent metrics come back null instead of failing. `low_signal_to_noise__bf_oblique` goes 10 -> 0 with 41 nulls, `low_tenengrad` 17 -> 12. Fewer flags, and the ones left mean something.
- **Per-entry `aggregate:`** in the report, which is what section 2 uses.
- **Flat plan v5** (`items[]` replacing `waves[]`) — the Nextflow driver was already adapted for this on #298.
- `report --config <file>` keeps the `report:` block (imaging-qc#201), which is what let the configs flatten to one file per store.

## 2. Default configs: max over Z for SNR and tenengrad, and tenengrad stops normalizing

Both `nextflow/configs/a549/qc.yaml` and `nextflow/configs/zebrafish/qc.yaml`.

**Max, not the default median.** The question a reader asks of a Z stack is "is there usable signal or focus *anywhere* in it". A median over 86 planes answers a different one — two thirds of this store's planes are out of focus, so it is dragged toward the floor. On GFP: median-over-Z SNR **12.0** against max-over-Z **26.7**.

**`tenengrad: normalize: false`,** which is what makes max-over-Z correct for it rather than actively wrong. Normalized, the metric divides gradient energy by an intensity scale, and in empty planes that divisor is near zero — so noise becomes the largest value in the stack. Measured on `2026_08_11_A549_SEC61B_DENV`:

| | peak Z | rho vs SNR across Z |
|---|---|---|
| `normalize: true` (before) | 3-19 — empty, SNR ~5 | **-0.31 to -0.69** |
| `normalize: false` (now) | 31-37 — the in-focus band, where SNR peaks | **+0.97 to +0.99** |

Harness verified first: recomputing with `normalize: true` reproduces the shipped table 2580/2580 identical, so the two runs differ only in that flag.

**No aggregation for anything else** — `intensity_stats` stays `range_drift`, and no `group_positions:` (see below).

## Known gap, deliberately deferred

⚠ **tenengrad's `hard_floor: 0.001` is now inert.** It is a normalized bound, and raw values run 179-934 on GFP — it cannot fire. Visible in the verdicts, where tenengrad passes with a summary value of 7623. Not guessed at here because the raw scale is per channel (BF near 2800, Phase3D near 1e-10, the predictions elsewhere again), so one scalar cannot serve six channels; it needs reading off healthy runs per channel, the way the SNR floors were. The relative outlier fence still gates the metric meanwhile.

## Notes

- `aggregate:` is **report-side only.** Gates still collapse Z by a hard-coded mean, so a panel and a verdict are two different reductions of the same rows. Asked upstream to gate by the configured reducer: imaging-qc#222.
- `group_positions:` omitted because it is report-*level*, so it would hit every panel at once — and the right value differs by form. `range_drift` is unreadable at 972 traces and wants per-well medians, while a facetgrid's distribution *is* the point and collapsing 54 fields to 6 destroys it (the GFP panel's max falls 21.8 -> 15.3). Requested as a per-entry key upstream: imaging-qc#221.
- Also filed upstream from this work: imaging-qc#220 (`metric_asrchetypes` typo silently accepted).

## Test

Full pipeline end to end on the A549 subset with the shipped configs: **11 tasks, 0 failed**, both report tabs, all 7 plots, and 0 unrecognized-key warnings.

### DIFF
--- a/nextflow/configs/a549/qc.yaml
+++ b/nextflow/configs/a549/qc.yaml
@@ -75,9 +75,23 @@ blank:
   gate:
     granularity: channel
 
+# NORMALIZE OFF, measured rather than preferred. With `normalize: true` this
+# metric divides gradient energy by an intensity scale, and in the empty planes
+# that divisor is near zero — so noise becomes the largest value in the stack.
+# On 2026_08_11_A549_SEC61B_DENV it peaked at Z 3-19 (SNR ~5 there, i.e. no
+# signal) and ANTI-correlated with SNR across Z, rho -0.31 to -0.69. Raw, on the
+# same pixels, it peaks at Z 31-37 — inside the in-focus band where SNR peaks —
+# and correlates at rho +0.97 to +0.99. It is a focus metric again.
+#
+# ⚠ `hard_floor` below is a NORMALIZED bound and is inert against raw values,
+# which run 179-934 on GFP: it cannot fire. Recalibration is deliberately
+# deferred — the floor is channel-scale dependent (BF, Phase3D and the
+# predictions each sit on their own scale), so it needs reading off healthy runs
+# per channel, like the SNR floors. The relative outlier fence still gates this
+# metric in the meantime.
 tenengrad:
   threshold: 0.0
-  normalize: true
+  normalize: false
   granularity: slice
   gate:
     hard_floor: 0.001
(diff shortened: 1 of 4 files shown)


ieivanov:
Nice to have:
- [ ] https://github.com/czbiohub-sf/imaging-qc-pipeline/issues/221
- [ ] https://github.com/czbiohub-sf/imaging-qc-pipeline/issues/222

aofei-liu (approved):

ieivanov:
@aofei-liu I'm going to go ahead and merge this, and we can bump the rev again when the issues above are closed. No big rush.

Issue #304, opened 2026-08-05
feat(nextflow): Restructure Nextflow DAG to work across positions rather than across reconstruction steps


ieivanov:
The current Nextflow pipeline runs recon steps in sequence for all positions, allowing for 30 jobs to run at a time. For example, if we have a store with 32 positions, we may run recon on the first 30 positions in parallel, and then have to wait for the recon of the other 2 positions to finish before moving on to the next recon steps. Instead, the Nextflow DAG can be restructured to work across positions - a given position can go through all recon steps, independent of the reconstruction of other positions.

A related change - init for all recon steps should run at the beginning, before the actual reconstruction work starts. This will allow us to validate recon configs and raise errors early rather than later on - for example, an error in the tracking config can surface several hours into the run.

It's possible that this feature will overcomplicate the DAG - it's worth checking if Nextflow has a way to batch such work units.

ieivanov:
Splitting this. The "related change" paragraph — init for all recon steps running at the beginning — is done separately in #334, since it stands on its own and is a prerequisite for the per-position DAG anyway (a position cannot stream through the steps until every output store has been scaffolded).

What's left here is the per-position restructuring, and two whole-plate steps block it.

### 1. `viscy preprocess` is the blocker at virtual staining

`run_virtual_stain_preprocess` runs `viscy preprocess` over the **whole** reconstruct store and mutates it with normalization statistics that per-position prediction reads back via `read_norm_meta`, erroring if they are missing (`nextflow/modules/virtual_stain.nf`, `biahub/virtual_stain.py`). It needs every position's reconstruction written, so no amount of channel rewiring gets a position from `reconstruct` into `virtual-stain` on its own.

Resolvable by computing the required normalization statistics **opt-in during reconstruction**. Virtual staining after phase reconstruction is a very common workflow, and the statistics are cheap to accumulate while the data is already in memory — far cheaper than the separate full-plate read pass we do now. That removes the preprocess step entirely and lets virtual staining stream per position.

### 2. `concatenate` is the second whole-plate barrier

Assemble runs single-shot on one reserved node, iterating every position in-process (`nextflow/modules/assembly.nf`). It should be refactored to work across positions, which also resolves #301.

### 3. Concurrency, once positions do stream

[...]

Issue #301, opened 2026-08-05
feat(nextflow): Develop faster Nextflow `concatenate` module


ieivanov:
In #279 I decided that `concatenate` should run on one mode without per-position fan-out. `concatenate` now has become the slowest step in our reconstruction pipeline. It would be useful to implement per-position fan-out as in #261 to speed up the overall pipeline execution. As an alternative, we could also test increasing the number of workers on one node, but that will likely not result in sufficient speed-up.

PR #334, opened 2026-09-12
feat(nextflow): run every step's init before any compute starts

ieivanov:
Addresses the "related change" paragraph of #304: *init for all recon steps should run at the beginning, before the actual reconstruction work starts.*

## The problem

Each step's `--init` was gated on the previous step's **compute** finishing, so a config was not parsed until the step that consumes it was about to run. A typo in `track.yml` surfaced seven hours in, with flat-field, deskew, reconstruction, virtual staining and assembly all already paid for.

## Why the inits can just be hoisted

Every `--init` is metadata-only. It reads the first input position's channel names, shape and scale, derives the output geometry from the config, creates the output plate (`create_empty_plate` is idempotent) and prints `RESOURCES:`. Nothing reads a pixel.

So a step's init needs its input store to **exist**, not to be **filled** — and the store that has to exist is the one the previous step's init just created. The inits chain:

```
init_flat_field → init_deskew → init_apply_inv_tf → init_virtual_stain
    → resolve_concatenate_config → init_concatenate → init_track
```

all on the head node, in the run's first minutes. It is real validation, not a schema check: concatenate resolves its channel mapping against all three source stores, virtual-stain validates its predict config against VisCy's schema, and every settings model is built with `extra="forbid"`.

## What changed

- Each step module splits into `<step>_init_wf` and `<step>_run_wf`. `mantis-v2.nf`, which already owned the order steps run in, now orders both phases, with one `init_done` gate between them: **nothing reaches SLURM until every config in the run has been parsed and every output store exists.**
- The per-step compute barriers are untouched. Per-position streaming is #304 proper and is blocked on `viscy preprocess` and single-shot `concatenate`, not on this wiring — see the issue comment.
- QC's `plan-stage` and `estimate-resources` join the init phase: both read only the store's structure and validate the config before opening it ("Neither reads a pixel" — imaging-qc `cli/planning.py`). `qc_stage_wf` becomes `qc_plan_wf` + `qc_compute_wf`, and each store's planned work is released by the step that filled it.
- The transfer function is built from the deskewed plate's **shape** alone, so `reconstruct_tf_wf` is gated on `init_done` rather than on deskew's data. It now overlaps flat-field and deskew instead of sitting between them and reconstruction — roughly 30 minutes off the critical path.

## The one thing that would have broken, and the fix

`create_empty_plate` inherits an upstream store's `biahub-*` zattrs **at plate-creation time and only for positions it creates**, while each step's own record was written per position at compute time. That works today only because a step's init runs after the previous step finished. Scaffold everything up front and the chain collapses — `1-deskew` would carry `biahub-deskew` and nothing else.

Fixed by stamping each step's record during `--init`, via a new `create_empty_plate(extra_metadata=...)` (czbiohub-sf/iohub#472). Putting it in `create_empty_plate` rather than in a pass afterwards costs no extra I/O — the function already holds an open handle to every position — and covers pre-existing positions, so a re-run with a changed config refreshes the record. The matching run-time write is dropped; `virtual-stain` and `track` already did this by hand and now go through the same seat.

czbiohub-sf/iohub#472 is **merged**; `iohub` is pinned to the resulting main
commit until a release carries it (`pyproject.toml`, `uv.lock`). Review there
scoped the reserved-key set to `_POSITION_OME_KEYS` and added OME-Zarr v0.4/v0.5
coverage.

## Dependency bumps riding along

- **iohub** → the merged #472 commit, for `create_empty_plate(extra_metadata=)`.
- **imaging-qc** → `c5c51dc` (imaging-qc#226), which makes `plan-stage` refuse a
  `channels:` or `per_channel_params` name no position in the store has. That
  closes the last QC config error the init phase could not catch: an unmatched
  filter used to be a legitimate zero, so the metric silently did not run and,
  if ungated, never surfaced. Raised off this pipeline's `qc_track.yaml`, whose
  `nuclei_prediction_labels` only matches if it agrees with `track.yml`'s
  `target_channel`. Verified against the v4 stores: both shipped a549 configs
  still plan 54 items at exit 0, and a mistyped channel is refused at exit 2.

## Verification

- `pytest tests` — 239 passed, 1 skipped.
- New `tests/test_cli/test_init_provenance.py`: two inits chained on a store **with no data** must still produce the full record. Both new assertions fail without this change (verified by reverting the flat-field stamp).
- DAG rendered under `-preview` and checked programmatically: all 9 init tasks are ancestors of all 9 compute tasks. Also builds for the neuromast (assemble + QC, no track), assemble-only, and reconstruction-only step sets.
- Full init chain run by hand against scaffolded, empty stores through `concatenate --init` and `track --init` (real `configs/a549/track.yml`, cellpose warm-up included). The provenance chain accumulates exactly as it does on a real run:
  ```
  1-deskew     ['biahub-deskew']
  2-recon      ['biahub-deskew', 'waveorder']
  3-vs         ['biahub-deskew', 'waveorder', 'cytoland']
  4-assemble   ['biahub-deskew', 'waveorder', 'cytoland', 'biahub-concatenate']
  5-track      ['biahub-deskew', 'waveorder', 'cytoland', 'biahub-concatenate', 'biahub-track']
  ```
  matching `2026_08_19_A549_PCNA_DENV_ZIKV` on disk.

## Full-scale run: done, and it found a bug

Ran all 54 positions of `2026_08_11_A549_SEC61B_DENV` end to end
(`.../2026_08_11_A549_SEC61B_DENV_debug/v4`), against v3 of the same data as the
reference. **The first launch failed** — see the gate-channel commit. Worth
stating plainly: neither the unit tests, nor `nextflow lint`, nor
`nextflow run -preview` catch that class of defect, because the failure is a
runtime arity mismatch that fires only when a gate channel emits. It cost all 54
positions of flat-field to surface. `nextflow/tests/gate_shapes.nf` now pins it.

Relaunched with `-resume` (flat-field came back entirely from cache) and
completed: **337 completed, 67 cached, 7 retried preemptions, 0 aborted.**

| phase | span |
|---|---|
| **init (9 tasks, head node)** | **2 min 32 s** |
| flat-field | ~37 min |
| deskew | ~20 min |
| reconstruct | ~27 min |
| virtual staining | ~2 h 40 m |
| assemble | 4 h 19 m |
| track | ~18 min |
| QC + report | ~1 h |

Every one of the nine init tasks completed before the first SLURM job was
submitted, including QC's `plan-stage`/`estimate-resources` — the latter planned
against `5-track`, a store that exists only because `track --init` scaffolded it,
two scaffold-only hops downstream of any data. `compute_transfer_function`
finished at 17:31, overlapping flat-field instead of waiting on 54 deskews.

**Provenance chain, the thing that would have broken.** Identical to v3 on all 54
positions at every step:

```
2-reconstruct  biahub-flat_field, biahub-deskew, waveorder
4-assemble     … + cytoland, biahub-concatenate
5-track        … + biahub-track
```

The `waveorder` block written at `--init` is byte-identical to v3's run-written
one, with a single `Phase3D` key — waveorder's merge overwrote in place rather
than duplicating. v3 additionally carries `normalization`; waveorder 3.0.5 (this
branch and main) never writes that key, so its absence is a dependency-version
difference, not a regression here — and it was already excluded from
`PROVENANCE_METADATA_KEYS`, so it never propagated downstream anyway.

**QC verdict, v4 vs v3:** assembled store `pass=37 fail=17` vs `pass=38 fail=16`;
tracking store `pass=44 fail=10` vs `pass=44 fail=10`. The one-position
difference on the assembled store is within the run-to-run variance the pipeline
already documents for segmentation.

Two observations that are not about this PR but came out of the run:

- assemble took **4 h 19 m** against ~2 h 15 m for v3 on the same data — it is
  single-shot on one reserved node, and it is the step #301 wants rewritten.
- track and QC run concurrently by design, so the run peaked at ~60 concurrent
  SLURM jobs. Today's pipeline therefore already exceeds a "30 total" cap
  independent of any per-position streaming; relevant to the concurrency point
  in #304.

## Note for anyone resuming an in-flight run

Task hashes change, so an output directory from an older biahub recomputes rather than resuming. biahub's own `--resume` still skips finished `(t, c)` units for flat-field, deskew and concatenate; `apply-inv-tf`, `virtual-stain` and `track` have no `--resume` and redo whole positions.

### DIFF
--- a/.claude/skills/reconstruct-with-nextflow/references/caveats.md
+++ b/.claude/skills/reconstruct-with-nextflow/references/caveats.md
@@ -291,6 +291,12 @@ plate for *all* positions with data in only the first — correct and expected.
 Do not treat such a store as a deliverable, and delete it before a real run so
 `-resume` cannot reuse it.
 
+Since the init phase moved to the front of the run, this is visible sooner and
+for every step at once: a smoke test creates the full-width plate for
+flat-field, deskew, reconstruct, virtual-stain, assemble AND track within the
+first minutes, before a single position has been computed. Nothing changed
+about what ends up on disk — only when it appears.
+
 (Related, for pipeline developers: a param passed as `--foo 1` on the command
 line arrives as a String; coerce numeric params at the point of use and test
 via the CLI — see the comments in `nextflow/modules/common.nf`.)
--- a/.claude/skills/reconstruct-with-nextflow/references/monitoring.md
+++ b/.claude/skills/reconstruct-with-nextflow/references/monitoring.md
@@ -122,10 +122,12 @@ Two traps, both hit for real:
 
 - **Name the store.** A bare `$d/*.zarr/*/*/*/` overcounts `2-reconstruct`, which
   also holds `transfer_function.zarr` (10 instead of 8 for an 8-position plate).
-- **This counts scaffolds, not finished work.** Each step's cached `init-*` task
-  creates every position's metadata up front, so a step shows its full position
-  count from the moment it starts — a step whose every task *failed* still reads
-  8/8 here. Use it for "has this step started", never for "is this step done".
+- **This counts scaffolds, not finished work.** Every step's `init_*` task runs
+  in the run's first minutes — the pipeline does the whole init chain before any
+  compute — and each creates every position's metadata up front. So EVERY step
+  reads its full position count from a few minutes into the run, including steps
+  that will not start for hours and steps whose every task later fails. Use this
+  only to confirm the init phase completed; it says nothing about progress.
   `trace.txt` is the only authoritative source for completion:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ nextflow/
 
 Each step module runs the same CLI you would run by hand, in two phases: `biahub <step> --init` creates the output store and reports the CPU/memory/time a single position needs, then Nextflow fans out one `biahub <step> --cluster debug` task per position with those resources. `--cluster debug` makes the CLI do the work in-process, so Nextflow — not `submitit` — owns job submission.
 
+**A run does every step's `--init` first, before any compute.** Each `--init` is metadata-only — it reads the input store's channel names, shape and scale, derives the output geometry from the config, and creates the output plate — so it needs its input store to exist, not to hold data. That lets the inits chain: flat-field's scaffolds the plate deskew's reads, and so on down to tracking, whose init reads the assembled plate `concatenate --init` created. The whole chain runs on the head node in the run's first minutes, so a typo in `track.yml` stops the run there instead of after every reconstruction step has been paid for. QC's `plan-stage` and `estimate-resources` join the same phase for the same reason.
+
 ### Running a pipeline
 
 Install [Nextflow](https://www.nextflow.io/docs/latest/install.html) (requires Java 17+), then launch from the `nextflow/` directory so that `nextflow.config` is picked up automatically:
--- a/biahub/apply_inverse_transfer_function.py
+++ b/biahub/apply_inverse_transfer_function.py
@@ -66,6 +66,15 @@ def _init_output_plate(
         **output_metadata,
         metadata_sources=input_plate,
         metadata_keys=PROVENANCE_METADATA_KEYS,
+        # The record waveorder itself writes once a position is reconstructed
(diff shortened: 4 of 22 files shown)


ieivanov:
Tested with `/hpc/projects/intracellular_dashboard/organelle_dynamics/2026_08_11_A549_SEC61B_DENV_debug/v4`. Waiting on https://github.com/czbiohub-sf/iohub/pull/472

PR #339, opened 2026-09-17
Robust per-timepoint beads registration: quality scoring, repair fallbacks, and votefit seed correction

tayllatheodoro:
## Summary

Hardens `estimate-stabilization --method beads` for long timelapses. The estimator gains
a per-timepoint quality score, a layered fallback system (repair → sweep → polish) for
timepoints where the independent estimate fails, and an opt-in per-timepoint seed
self-correction ("votefit") that fixes whole-region failures no local fallback can reach.
Defaults are chosen so that existing configs run unchanged and bit-identically; every new
behavior is opt-in or triggered only where the previous behavior produced a failed
timepoint.

## Motivation

Validated across an 11-dataset zebrafish neuromast campaign (mantis, 144–848 timepoints
each). The failure modes this PR addresses were all observed in production:

- **Stale seeds**: a config `approx_transform` copied across acquisition epochs lands
  ~55+ voxels off; the matcher cannot bridge it and a whole dataset collapses
  (three datasets, medians 0.00–0.12 before / 0.4–0.8 after).
- **Mid-series drift and O3 excursions**: single timepoints or short runs where the
  stage/optics move faster than the seed can track (validated rescues 0.00 → 0.81–0.89).
- **Restart clobbering**: resubmitting the independent-estimate driver overwrote
  already-repaired timepoints and the journal resume then skipped them forever (227
  timepoints on one dataset).
- **Silent score staleness**: repair never persisted the score sidecar, so `.score`
  files could contradict the transforms on disk.

## What's in the change

**Scoring & QC**
- Per-timepoint quality score persisted as a `.score` sidecar and plotted; QC summary
  written as csv alongside the json.
- Flagging recalibrated; plausibility QC replaced by score gates.

**Fallback system** (runs only on flagged/failed timepoints)
- *Repair*: reseed a flagged timepoint from both neighbours, screen the flagged
  translation before reuse, seed repairs from consensus geometry.
- *Sweep*: score-gated per-timepoint parameter sweep, restrictable to the repair pass's
  failures; both fallbacks can run in competition (all five combinations exposed).
- *Polish*: an accepted repair re-seeds the cascade; pre-polish score recorded so each
  step is attributable; raised round cap for the worst timepoints.
- Config restructured under one `fallback_settings` block (default: repair alone).

**Votefit seed correction** (opt-in: `beads_match_settings.seed_correction_settings.mode: votefit`)
- Per-timepoint seed correction by bead-displacement voting; fixes region-wide seed
  drift that neighbour-reseeding cannot (signature: many repair attempts, failing
  fraction unchanged). Default `none` is bit-identical to current behavior.

**Robustness fixes**
- Restart no longer clobbers repaired timepoints (skip-existing on the independent path).
- Survive timepoints with too few beads; resume support; propagate the last good
  transform instead of the config seed; repaired transforms no longer break the settings
  yml; `edge_graph_settings.method` honoured; spectral matching exposed as an opt-in
  three-way cascade.

## Validation

- 11/11 campaign datasets registered end-to-end with this branch pinned; beads-score
  medians 0.72–0.83 on the hard datasets, zero timepoints below 0.2 after votefit.
- Votefit validated bit-for-bit against a parameter sweep on one dataset
  (0.000 → 0.545/0.385 on failed timepoints, good timepoints preserved).
- Bad-seed diagnosis and per-dataset consensus-seed recovery documented per dataset.
- Default-path regression: with all new settings at defaults, transforms are
  bit-identical to the pre-branch estimator on a reference dataset.

## Notes for reviewers

- `f30f354` duplicates the already-merged #296 (AntsRegistrationSettings fix) and will
  vanish on merge.
- Not included (stacked branches, pending their own validation): vote_icp, single-t
  touch-up, bootstrap/CPD, escalation loop.

### DIFF
--- a/biahub/core/graph_matching.py
+++ b/biahub/core/graph_matching.py
@@ -289,7 +289,7 @@ class GraphMatcher:
 
     def __init__(
         self,
-        algorithm: Literal["hungarian", "descriptor"] = "hungarian",
+        algorithm: Literal["hungarian", "descriptor", "spectral"] = "hungarian",
         weights: dict[str, float] | None = None,
         distance_metric: str = "euclidean",
         normalize: bool = False,
@@ -297,6 +297,9 @@ class GraphMatcher:
         cross_check: bool = False,
         max_ratio: float | None = None,
         metric: str = "euclidean",  # for descriptor matching
+        spectral_sigma: float = 3.0,
+        spectral_rel_cut: float = 0.5,
+        spectral_max_iter: int = 60,
         verbose: bool = False,
     ):
         self.algorithm = algorithm
@@ -314,6 +317,11 @@ class GraphMatcher:
         self.distance_metric = distance_metric
         self.normalize = normalize
         self.cost_threshold = cost_threshold
+
+        # Spectral-specific parameters
+        self.spectral_sigma = spectral_sigma
+        self.spectral_rel_cut = spectral_rel_cut
+        self.spectral_max_iter = spectral_max_iter
 
         # Common parameters
         self.cross_check = cross_check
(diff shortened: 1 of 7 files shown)


ieivanov:
## `estimate-stabilization` CLI changes in 3687140 — practical consequences

Commit 3687140 (on top of the cosmetic/type-hint cleanup in ded57de) makes four user-visible changes to `biahub estimate-stabilization`. Summary of what changes for people running it:

### 1. `--local` is replaced by `--cluster {slurm,local,debug}`
- **Breaking:** any script, Nextflow step, or shell history that calls `biahub estimate-stabilization --local` will now fail with "no such option". Use `--cluster local` instead. The default (`--cluster slurm`) is unchanged, so plain SLURM invocations are unaffected.
- **New:** `--cluster debug` runs every per-position job in-process in the foreground. This is what CI now uses (the `CI=true` env var forces debug), and it lets a Nextflow worker run one position without spawning submitit subprocesses. I verified this end to end on a synthetic two-position plate.
- `estimate-registration` and `stabilize` still use the legacy `--local` flag; migrating them is a follow-up.

### 2. SLURM memory request and job-ID log
- Memory is now requested as `slurm_mem = cpus × RAM-per-CPU` (matching `deskew` and the other processing commands) instead of `slurm_mem_per_cpu`. The total is the same, only the SBATCH directive differs, so an `--sbatch-filepath` override that sets `--mem-per-cpu` will still take effect via `sbatch_to_submitit`.
- Job IDs now go to `<output>/slurm_output/submitit_jobs_ids.log` (same filename as every other biahub command) instead of `job_ids_<timestamp>.log`. The file is **appended**, because the XYZ focus-finding path submits two stages (Z focus, then XY stack-reg) into the same folder. If you re-run into an existing output directory, the log will contain IDs from previous runs too.

### 3. Provenance file
- The command now writes `<output>/estimate_stabilization_settings.yml` containing the fully resolved `EstimateStabilizationSettings` it ran with (including defaults not present in your input config). This is the counterpart of the `biahub-<command>` metadata block the processing commands write into zarr attrs, and it lets you reproduce or audit which estimation settings produced a given set of `*_stabilization_settings/*.yml` transforms.

### 4. Failures now exit non-zero
- Previously, any error during transform evaluation or saving (for example `evaluate_transforms` refusing to run because there are fewer timepoints than `validation_window_size`) was caught, echoed, and swallowed. The command exited 0 with no `*_stabilization_settings/` output, and a downstream `stabilize` or Nextflow step would fail later on a missing input, far from the cause.
- Now the same message is echoed and the exception is re-raised, so the command exits non-zero with a traceback. **If you have pipelines that tolerated a "successful" estimate step with missing outputs, they will now stop at the estimate step instead.** That is the intended behaviour, but worth knowing when comparing against older runs.

ieivanov:
@tayllatheodoro along with your changes I also added a bit of cleanup to `biahub/estimate_stabilization.py` to align it with the style of the rest of the CLIs using the `biahub-cli-consistency` skill. Please take a look at [3687140](https://github.com/czbiohub-sf/biahub/pull/339/commits/36871407227d9997ddc0ba9f79b5507cb91ba8ad) to make sure it wouldn't introduce any breaking changes for you. Otherwise, I think this PR is good to merge.

ieivanov (approved):

PR #341, opened 2026-09-17
fix(nextflow): retries ask for double the base resources, and no more

ieivanov:
## Summary

Every fan-out process scaled its `time` by `task.attempt` (flat-field, deskew, reconstruct, virtual-stain, assemble), and `run_virtual_stain_preprocess` and the QC `compute_step` / `finalize_stage` scaled `memory` the same way. A retried task therefore asked for 2x, then 3x, … the estimate. But a retry here is almost always a **preemption** — 5 of the 6 retries in the full 2026_08_11 v5 run, exit 143 or an unreadable status — which says nothing about the task's size. Two consequences:

- a twice-preempted position waited for a triple-size slot it did not need;
- above the `preempted` partition's 48 h limit, sbatch rejects the request with exit 1, which the errorStrategy does not retry, so any estimate over 24 h killed the whole run on its first preemption. That is how `2026_08_14_dynatrack` died this morning (flat-field base 27 h → 54 h on attempt 2; see the `nextflow_walltime_override.config` the operator staged next to that run).

Two helpers in `modules/common.nf` replace the seven directives:

```groovy
time   { retry_time(meta.time_minutes, task) }   // base, then 2x base; capped at 2880 min
memory { retry_memory(64, task) }                // base, then 2x base
```

A retry asks for **double the base, and no more on later attempts**. The one behavioural change besides that: QC `finalize_stage` retries with 64 GB instead of 48 GB.

## Testing

- New `nextflow/tests/retry_resources.nf`, runnable in seconds like `gate_shapes.nf`: asserts base / 2x / 2x / 2x across attempts 1, 2, 3, 6 for time and memory, the 2880 min cap for a doubled 27 h base and for a base already above the limit, and — through a process that exits 143 on its first two attempts — that the third attempt really runs with `time=20 memory=4` from a base of 10 min / 2 GB.
- `nextflow/tests/gate_shapes.nf` still passes; `nextflow run mantis-v2.nf -preview` with every optional step on compiles and succeeds (26.04.3).

Independent of #337 and #340 (touches the same modules on different lines; trivial rebases either way).


ieivanov:
Closing in favor of #342

